PR #955, opened 2024-07-25
Chunked generation in OpenAI API server

vmpuri:
To test:
```
python3 torchchat.py server stories15M
```

In another terminal, send request with "stream"="true":
```
curl http://127.0.0.1:5000/chat \
  -d '{
    "model": "stories15M",
    "stream": "true",
    "messages": [
      {
        "role": "system",
      },
      {
        "role": "user",
      }
    ]
  }'
```
https://github.com/user-attachments/assets/590d8cb4-075e-43ed-8849-2490a92f2e9f


README.md
```diff
@@ -186,6 +186,33 @@ curl http://127.0.0.1:5000/chat \
...
 ```
 
+In the request, setting `stream` to "true" emits a response in chunks. Currently, this response
+is plaintext and not formatted to the OpenAI API specification.
+```
+curl http://127.0.0.1:5000/chat \
+  -H "Content-Type: application/json" \
+  -d '{
+    "model": "stories15M",
+    "stream": "true",
```
Jack-Khuu, line 196: Can we do this with llama3 and collapse the 2 code snippets?

So 
<Curl with stream=false>
<output with false>
> In the request, setting `stream` to "true" emits a response in chunks. Currently, this response
is plaintext and not formatted to the OpenAI API specification
<output with true>

Jack-Khuu (approved):

README.md
vmpuri: The output would be exactly the same whether streamed or not. For now I"ll just remove both new code blocks and just set stream explicitly in the arguments.

PR #961, opened 2024-07-29
rename model args into transformer args

Gasoonjia:
as title


Jack-Khuu:
Did you want to land this rename first? If so, can you update the title of the PR, since this only a component of it

Jack-Khuu (approved):

PR #962, opened 2024-07-29
README: Add a model customization guide

Jack-Khuu:
This adds an initial index for model customization options accessible from the README.

The introduced a "model_customization.md" that will be iterated and expanded upon over time

Note that some of the content is extracted/inspired by quantization.md and the ADVANCED_USERS docs, which are marked as outdated/unstable.

**README's**
https://github.com/pytorch/torchchat/blob/readme-model-customization-guide/README.md
https://github.com/pytorch/torchchat/blob/readme-model-customization-guide/docs/model_customization.md

### DIFF
--- a/docs/model_customization.md
+++ b/docs/model_customization.md
@@ -1,0 +1,60 @@
+# Model Customization
+
+By default, torchchat (and PyTorch) default to unquantized [eager execution](https://pytorch.org/blog/optimizing-production-pytorch-performance-with-graph-transformations/).
+
+This page goes over the different options torchchat provides for customizing the model upon which inference is done.
+- Compilation
+- Quantization
+- Model Precision
+- Device
+
+
+## Compilation: JIT-compiled execution
+```
+python3 (chat | generate | browser | server | eval) [--compile] ...
+```
+
+To improve performance, you can compile the model with `--compile`;
+trading off the time to first token processed with time per token.
+
+To improve performance further, you may also compile the prefill with
+`--compile_prefill`. This will increase further compilation times though.
+
+To learn more about compilation, check out: https://pytorch.org/get-started/pytorch-2.0/
+
+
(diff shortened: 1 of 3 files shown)


Gasoonjia (approved):

docs/model_customization.md
```diff
@@ -0,0 +1,61 @@
+# Model Customization
+
+By default, torchchat (and PyTorch) defaults to unquantized [eager execution](https://pytorch.org/blog/optimizing-production-pytorch-performance-with-graph-transformations/).
+
+This page goes over the different options torchchat provides for customizing the model execution for inference.
+- Device
+- Compilation
```
digantdesai, line 7: This doesn't fit very well IMHO in this otherwise awesome high level, almost orthogonal optimization knob categorization. It could be just me.

digantdesai (approved):
Love it. For compile, it is not really a model customization, if we can, to be more complete, extend that category of this doc to encompass that to `(et | aoti | eager | compile)` it might be more comprehensive / useful.

Also It would be great if you have a "support matrix" view of these four knobs for a flagship model for torchchat which would be llama3-8b I suppose.

docs/model_customization.md
```diff
@@ -0,0 +1,60 @@
+# Model Customization
+
+By default, torchchat (and PyTorch) default to unquantized [eager execution](https://pytorch.org/blog/optimizing-production-pytorch-performance-with-graph-transformations/).
```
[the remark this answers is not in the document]
Jack-Khuu: It's the checkpoint dtype, so it'll vary

docs/model_customization.md
Jack-Khuu: Agreed, it's not an optimization knob, it's a model customization knob

Quant/optimization gets it's own page

PR #974, opened 2024-07-30
[Draft Test] CI: Set the MPS watermark to avoid running out of MPS memory

Jack-Khuu:
Some CI tests using MPS were failing due to memory usage. This explicitly sets the RATIO as suggested
```
Please file an issue with the following so that we can make `weights_only=True` compatible with your use case: WeightsUnpickler error: MPS backend out of memory (MPS allocated: 1.02 GB, other allocations: 0 bytes, max allowed: 15.87 GB). Tried to allocate 256 bytes on shared pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 to disable upper limit for memory allocations (may cause system failure).
```

**Example Job:** https://github.com/pytorch/torchchat/actions/runs/10166476129/job/28116731017

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -449,6 +449,7 @@ jobs:
           fi
       - name: Install requirements
         run: |
+          PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0
           echo "Intalling pip3 packages"
           ./install_requirements.sh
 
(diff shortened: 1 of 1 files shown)


byjlw:
@Jack-Khuu won't this cause other issues though?

Jack-Khuu:
Just testing things out, this doesn't fix the CI (same error that it tries to address shows up)

Issue #979, opened 2024-07-31
Self-documenting repo: Too many files and folders in root directory


soumith:
### 🚀 The feature, motivation and pitch

The repo quite doesn't pass the smell test for being an example in that the number of files and folders in root are too many.
A lot of times, repos are self-documenting, by keeping the file/folder structure simple.

I couldn't find ton of scope for files, but I think:

* I think `requirements-lintrunner.txt` can go somewhere else.
* It was annoying that `export.py` literally had no export logic and was pointing to a full folder of files.
  * I think `export.py` should just have two functions, `export_edge` and `export_server`, and all logic we need should be in-lined there, instead of the reader having to jump across multiple files. Also, the `export_edge` code should be simplified (right now its across 3 or 4 files, not sure what its doing).

The folders are just too many. We should try to get down to 4 root folders that make sense?

PR #983, opened 2024-07-31
Move export_aoti into export.py

Jack-Khuu:
Flattening the export files per suggestion in https://github.com/pytorch/torchchat/issues/979

Specifically this moves the AOTI export logic into export.py (with some added docstring + arg types)
- Also removed unused arg in signature


byjlw:
since you're refactoring here anyway. Why not kill another top level directory and move this under the utils directory?

Jack-Khuu:
Let's have those directory moving PR's be separate

byjlw (approved):

PR #995, opened 2024-08-01
OpenAI API JSON formatted 

vmpuri:
Implement JSON formatted responses using OpenAI API types for server completion requests. Rather than giving single tokens at a time, the server will respond with a JSON following the API dataclasses corresponding to OpenAI API types. 

Testing:
Server 
```
python3 torchchat.py server stories15M   
```

Request (chunked)
```
curl http://127.0.0.1:5000/chat \
  -d '{
    "model": "stories15M",
    "stream": "true",
    "messages": [
      {
        "role": "system",
      },
      {
        "role": "user",
      }
    ]
```

Request (synchronous)
```
curl http://127.0.0.1:5000/chat \
  -d '{
    "model": "stories15M",
    "stream": "false",
    "messages": [
      {
        "role": "system",
      },
      {
        "role": "user",
      }
    ]
```

https://github.com/user-attachments/assets/1171db63-094b-4313-a1c6-76d042ceefca

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -8,7 +8,7 @@ import time
 import uuid
 from abc import ABC
 from dataclasses import dataclass
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Union
 
 from build.utils import device_sync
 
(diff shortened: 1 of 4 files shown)


Jack-Khuu:
Can you add a video for non-chunked as well for record purposes

api/api.py
```diff
@@ -220,10 +219,24 @@ def __init__(self, *args, **kwargs):
             if self.draft_model is not None
             else self.model.config.max_seq_length
         )
+        self.system_fingerprint = (
+            self.builder_args.device + type(self.builder_args.precision).__name__
+        )
```
Jack-Khuu, line 224: Let's add a comment that this field doesn't match the spec, but is populated

We'll fix in a separate PR

server.py
```diff
@@ -4,62 +4,83 @@
...
+"""
+Creates a flask app that can be used to serve the model as a chat API.
+"""
 app = Flask(__name__)
 # Messages and gen are kept global so they can be accessed by the flask app endpoints.
 messages: list = []
 gen: OpenAiApiGenerator = None
 
 
+def _del_none(d: dict):
```
Jack-Khuu, line 27: use `from typing ...` lib instead of the Python dict for typehint

But beyond that the typing seems off based on your function: it can be an dict or list

server.py
```diff
@@ -4,62 +4,83 @@
...
+
+        def chunk_processor(chunked_completion_generator):
+            """Inline function for postprocessing CompletionResponseChunk objects.
+
+            Here, we just jsonify the chunk and yield it as a string.
+            """
+            messages.append(AssistantMessage(content=""))
+            for chunk in chunked_completion_generator:
+                nextok = chunk.choices[0].delta.content
+                nextok = nextok if nextok is not None else ""
```
Jack-Khuu, line 69:
```suggestion
                if (next_tok := chunk.choices[0].delta.content) is None:
                   next_tok = ""
```

Jack-Khuu (approved):
Some nits, but the content lgtm

Remember to verify the behavior lines up with how people plan on using OpenAI

server.py
vmpuri: fixed

byjlw (approved):

Issue #1003, opened 2024-08-03
No attribute 'prompt' and 'num_samples'


YeonwooSung:
When running the server by `python3 torchchat.py server llama3.1`, I was facing with the following error:

```
Traceback (most recent call last):
  File "/Users/ywsung/OpenSources/torchchat/torchchat.py", line 84, in <module>
    server_main(args)
  File "/Users/ywsung/OpenSources/torchchat/server.py", line 85, in main
    gen = initialize_generator(args)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ywsung/OpenSources/torchchat/server.py", line 69, in initialize_generator
    generator_args = GeneratorArgs.from_args(args)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ywsung/OpenSources/torchchat/generate.py", line 113, in from_args
    prompt=args.prompt,
           ^^^^^^^^^^^
AttributeError: 'Namespace' object has no attribute 'prompt'
```

The reason of the error seems like that the "generate config" requires prompt.
And as you could see below, the server script internally initializes the GeneratorArgs instance by calling `GeneratorArgs.from_args(args)`:
```
# Add CLI Args specific to user prompted generation
# Include prompt and num_sample args when the subcommand is generate
def _add_generation_args(parser, verb: str) -> None:
    generator_parser = parser.add_argument_group(
        "Generation", "Configs for generating output based on provided prompt"
    )

    if verb == "generate":
        generator_parser.add_argument(
            "--prompt",
            type=str,
            default="Hello, my name is",
            help="Input prompt for manual output generation",
        )
        generator_parser.add_argument(
            "--num-samples",
            type=int,
            default=1,
            help="Number of samples",
        )
```

Also, it not only 'prompt' but also 'num_samples' is empty..

Jack-Khuu:
Ah looks like I missed this in https://github.com/pytorch/torchchat/pull/987

Thanks for spinning up a fix, I’ll comment there

Jack-Khuu:
Landed fix, thanks again for flaggin

Issue #1002, opened 2024-08-03
No attribute Prompt


bookandlover:
### 🐛 Describe the bug

AttributeError: 'Namespace' object has no attribute 'prompt'

### Versions

python collect_env.py

Collecting environment information...
PyTorch version: 2.5.0.dev20240710
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 15.0 (arm64)
GCC version: Could not collect
Clang version: 14.0.3 (clang-1403.0.22.14.1)
CMake version: version 3.30.1
Libc version: N/A

Python version: 3.12.4 (main, Jun  6 2024, 18:26:44) [Clang 15.0.0 (clang-1500.3.9.4)] (64-bit runtime)
Python platform: macOS-15.0-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M2 Max

Versions of relevant libraries:
[pip3] numpy==1.26.4
[pip3] torch==2.5.0.dev20240710
[pip3] torchao==0.3.1
[conda] Could not collect

YeonwooSung:
According to the `cli.py` file, the argument 'prompt' is only added for "generate", but the initializer of the server requires the prompt

```python
# Add CLI Args specific to user prompted generation
# Include prompt and num_sample args when the subcommand is generate
def _add_generation_args(parser, verb: str) -> None:
    generator_parser = parser.add_argument_group(
        "Generation", "Configs for generating output based on provided prompt"
    )

    if verb == "generate":
        generator_parser.add_argument(
            "--prompt",
            type=str,
            default="Hello, my name is",
            help="Input prompt for manual output generation",
        )
        generator_parser.add_argument(
            "--num-samples",
            type=int,
            default=1,
            help="Number of samples",
        )
```

Jack-Khuu:
Ah looks like i missed this in https://github.com/pytorch/torchchat/pull/987

Thanks for flagging

Jack-Khuu:
Merging with https://github.com/pytorch/torchchat/issues/1003

thanks again for flagging

PR #1004, opened 2024-08-03
Fix for "No attribute 'prompt' and 'num_samples'" issue

YeonwooSung:
When running the server by `python3 torchchat.py server llama3.1`, application throws the attribution error.

This PR fixes these issues:
#1002 #1003

### DIFF
--- a/server.py
+++ b/server.py
@@ -64,6 +64,12 @@ def chat_endpoint():
 
 def initialize_generator(args) -> OpenAiApiGenerator:
     builder_args = BuilderArgs.from_args(args)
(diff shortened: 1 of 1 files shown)


Jack-Khuu:
Thanks for flagging and spinning this up!!

I’m not at my laptop right now, but the idea looks good! Do you mind making the change in GeneratorArgs instead? That way it reflects in subcommands that use it

Jack-Khuu:
Thanks for suggesting a fix, it's been patched in https://github.com/pytorch/torchchat/pull/1007

PR #987, opened 2024-08-01
CLI: Fix unsafe arg access of unused args

Jack-Khuu:
As described in Issue [932](https://github.com/pytorch/torchchat/issues/932), the legacy implementation of the arg parser results in subcommands requiring the existence of cli args that they don't actually use.

This PR fixes this by doing a safe getattr check instead of a raw field access.

As a side effect, it also allows us to remove conditional suppression of args. 
Instead we can just omit the arg from the parser if they aren't needed.

Also happens to solve a --help bug https://github.com/pytorch/torchchat/issues/976

---
<details>
<summary>chat</summary>

```
 python torchchat.py chat --help
usage: torchchat chat [-h] [--checkpoint-path CHECKPOINT_PATH] [--compile] [--compile-prefill] [--dtype {fp32,fp16,bf16,float,half,float32,float16,bfloat16,fast,fast16}] [--quantize QUANTIZE] [--device {fast,cpu,cuda,mps}]
                      [--dso-path DSO_PATH | --pte-path PTE_PATH] [--max-new-tokens MAX_NEW_TOKENS] [--top-k TOP_K] [--temperature TEMPERATURE] [--hf-token HF_TOKEN] [--model-directory MODEL_DIRECTORY] [-v] [--seed SEED]
                      [model]

options:
  -h, --help            show this help message and exit
  -v, --verbose         Verbose output
  --seed SEED           Initialize torch seed

Model Specification:
  (REQUIRED) Specify the base model. Args are mutually exclusive.

  model                 Model name for well-known models
  --checkpoint-path CHECKPOINT_PATH
                        Use the specified model checkpoint path

Model Configuration:
  Specify model configurations

  --compile             Whether to compile the model with torch.compile
  --compile-prefill     Whether to compile the prefill. Improves prefill perf, but has higher compile times.
  --dtype {fp32,fp16,bf16,float,half,float32,float16,bfloat16,fast,fast16}
                        Override the dtype of the model (default is the checkpoint dtype). Options: bf16, fp16, fp32, fast16, fast
  --quantize QUANTIZE   Quantization options. pass in as '{"<mode>" : {"<argname1>" : <argval1>, "<argname2>" : <argval2>,...},}' modes are: embedding, linear:int8, linear:int4, linear:a8w4dq, precision.
  --device {fast,cpu,cuda,mps}
                        Hardware device to use. Options: cpu, cuda, mps

Exported Model Path:
  Specify the path of the exported model files to ingest

[...]

PR #1007, opened 2024-08-03
Fix GeneratorArgs to avoid unsafely accessing prompt and num_samples

Jack-Khuu:
Prompt and num_samples were restricted from suppressed arg (always accepted as CLI arg even when not used by subcommand) to being restricted to `generate` in https://github.com/pytorch/torchchat/pull/987.

But this PR missed a call site. This updates the callsite that was missed to not access the field without checking for existence

Tested that both these commands no longer crash with accessing unavailable fields:
- python3 torchchat.py generate llama3.1
- python3 torchchat.py server llama3.1

Fixes: https://github.com/pytorch/torchchat/issues/1003

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -110,11 +110,11 @@ class GeneratorArgs:
         )
 
         return cls(
-            prompt=args.prompt,
+            prompt=getattr(args, "prompt", ""),
             encoded_prompt=None,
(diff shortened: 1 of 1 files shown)


Jack-Khuu:
Force land to fix branch

PR #1011, opened 2024-08-05
[AOTI] Add a --dynamic-shapes option to export

desertfire:
Summary: The --dynamic-shapes option will default to False. When the actual inputs are with static shapes, calling export with static shapes will make sure more Inductor optimizations take effect down the line. This change by itself improves average tokens/sec from 29.60 to 33.43 on A100. Some following PRs will provide further perf gains.

```
python3 torchchat.py export llama3 --quantize '{"precision": {"dtype":"bfloat16"}, "executor":{"accelerator":"cuda"}}' --output-dso-path /tmp/model16.so && python3 torchchat.py generate llama3 --dso-path /tmp/model16.so --prompt "Once upon a time," --max-new-tokens 256 --device cuda --num-samples 3
```

### DIFF
--- a/.ci/scripts/validate.sh
+++ b/.ci/scripts/validate.sh
@@ -285,7 +285,7 @@ function eval_model_sanity_check() {
             echo "******** INT4 group-wise quantized (AOTI) *******"
             echo "*************************************************"
             if [ "$DTYPE" != "float16" ]; then
-                python3 -W ignore export.py --dtype ${DTYPE} --quant "$QUANT_OPTIONS" --checkpoint-path "$CHECKPOINT_PATH" --output-dso-path ${MODEL_DIR}/${MODEL_NAME}.so --device "$TARGET_DEVICE" || exit 1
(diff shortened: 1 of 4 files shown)


Jack-Khuu:
Great to see the perf gains, is anything lost by switching over to static? Does model support change?

desertfire:
> Great to see the perf gains, is anything lost by switching over to static? Does model support change?

ET export is doing the same so it should be fine, https://github.com/pytorch/torchchat/blob/912917f8afca9361d46bac50949e1353861353a8/export_util/export_et.py#L57-L65

Jack-Khuu:
Amazing, thanks for adding this

Jack-Khuu (approved):

desertfire:
> Amazing, thanks for fixing this
> 
> One last ask: Can you put the repro commands for the #'s in your description?

Done. We should add a benchmarking script so that everyone can run the same experiment.

PR #1016, opened 2024-08-06
Openai api completion params [seed, temperature, max_tokens, system_fingerprint]

vmpuri:
**Implement [seed, temperature, max_tokens, system_fingerprint(partial)]** 
Improve determinism of generated responses. 
- seed: simply set torch.manual_seed() each time the seed is provided
- max_tokens: default to 16 as per API spec.
- temperature: default to 1.0 as per API spec.  
- system_fingerprint: for now, the backend and torch.dtype of the model weights. In the future, we should add model version information to this. 

Testing: With the same seeds, the generated response is the same. 

https://github.com/user-attachments/assets/7b186f86-5855-4b9e-92c3-cc617727c752

Tested with the following cURL:
```
curl http://127.0.0.1:5000/v1/chat \
  -d '{
    "model": "stories15M",
    "stream": "true",
    "max_tokens": "123",
    "temperature": "0.2",
    "seed": "813",
    "messages": [
      {
        "role": "system",
      },
      {
        "role": "user",
      }
    ]
  }'
```

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -19,6 +19,8 @@ from generate import Generator, GeneratorArgs
 See https://platform.openai.com/docs/api-reference/chat for the full specification and details.
 """
 
+OPENAI_API_DEFAULT_MAX_TOKENS = 16
+
 # Message classes and associated objects - see the types of Messages under "Create Chat Completion >>> Request body >>> messages"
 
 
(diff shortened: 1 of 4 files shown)


api/api.py
```diff
@@ -229,9 +229,8 @@ def __init__(self, *args, **kwargs):
             else self.model.config.max_seq_length
         )
         # The System fingerprint is a unique identifier for the model and its configuration.
-        # Currently, this is not implemented in a
         self.system_fingerprint = (
-            self.builder_args.device + type(self.builder_args.precision).__name__
+            self.builder_args.device + "_" + str(self.builder_args.precision)
```
byjlw, line 233: f"{self.builder_args.device}_{self.builder_args.precision}"

api/api.py
```diff
@@ -270,7 +269,11 @@ def chunked_completion(self, completion_request: CompletionRequest):
         )
         generator_args = GeneratorArgs(
             completion_request.messages[-1].get("content"),
+            max_new_tokens=(
+                completion_request.max_tokens if completion_request.max_tokens else 16
```
byjlw, line 273: Define DEFAULT_MAX_TOKENS at the top and set it to 16 then
max_new_tokens = completion_request.max_tokens or DEFAULT_MAX_TOKENS

generate.py
```diff
@@ -71,7 +71,7 @@ class GeneratorArgs:
     num_samples: int = 1
     max_new_tokens: int = 200
     top_k: int = 200
-    temperature: int = 0  # deterministic argmax
+    temperature: float = 0  # deterministic argmax if 0.0
```
byjlw, line 74: temperature: float = 0.0

byjlw (changes requested):
several mostly nits other than that looks good

PR #1018, opened 2024-08-07
[AOTI] Add a --max-seq-length option for export

desertfire:
Summary: This improves best tokens/sec from 73 to 85.


cli.py
```diff
@@ -185,11 +186,20 @@ def _add_export_output_path_args(parser) -> None:
...
     parser.add_argument(
         "--dynamic-shapes",
         action="store_true",
         help="Call torch.export with dynamic shapes",
     )
+    parser.add_argument(
+        "--max-seq-length",
+        type=int,
+        default=None,
+        help="Set maximum length sequence when before calling torch.export",
```
byjlw, line 201: set the default to 300 and update the help string so that it's clear what the default is (300)

export.py
```diff
@@ -113,10 +113,19 @@ def main(args):
         except:
             tokenizer = None
 
+        if (
+            output_dso_path is not None
+            and builder_args.max_seq_length is None
+            and not builder_args.dynamic_shapes
+        ):
+            print("Setting max_seq_length to 300 for DSO export.")
+            builder_args.max_seq_length = 300
```
byjlw, line 122: you shouldn't need this if you set the default in the other file

byjlw (dismissed):

cli.py
desertfire: That was my initial implementation. Then there was an issue when running eval. 

When running eval, we use `--dynamic-shapes` which uses a larger max_seq_length, i.e. model.config.max_seq_length. But in theory, we should not stop user from calling eval with both options, something like `--dynamic-shapes --max-seq-length 1000`.  When that happens, if args.max_seq_length has a default value, we will have no way to distinguish if args.max_seq_length is from a default value or from an intentional user overwriting.

export.py
desertfire: I can add more details to the printout if that helps.

Jack-Khuu (approved):

PR #1021, opened 2024-08-08
Merge Openai api version route to main

vmpuri:
https://github.com/pytorch/torchchat/pull/1016 mistakenly got merged into this development branch instead of main.


README.md
```diff
@@ -200,6 +200,10 @@ streamlit run torchchat.py -- browser llama3.1
 <details>
 <summary>This mode gives a REST API that matches the OpenAI API spec for interacting with a model</summary>
 
+The server follows the [OpenAI API specification](https://platform.openai.com/docs/api-reference/chat) for chat completions.
+Since this feature is under active development, it's possible not every parameter is consumed. See api/api.py for details on
```
Jack-Khuu, line 204:
```suggestion
Since this feature is under active development, not every parameter is consumed. See api/api.py for details on
```

api/api.py
```diff
@@ -270,7 +271,13 @@ def chunked_completion(self, completion_request: CompletionRequest):
         )
         generator_args = GeneratorArgs(
             completion_request.messages[-1].get("content"),
+            max_new_tokens=(
+                int(completion_request.max_tokens)
```
Jack-Khuu, line 275: why cast?

api/api.py
```diff
@@ -270,7 +271,13 @@ def chunked_completion(self, completion_request: CompletionRequest):
...
+                if completion_request.max_tokens
+                else OPENAI_API_DEFAULT_MAX_TOKENS
+            ),
             encoded_prompt=encoded,
+            temperature=float(completion_request.temperature),
```
Jack-Khuu, line 280: ditto

Jack-Khuu (approved):

api/api.py
vmpuri: The request object was being populated from JSON, which automatically means the value for every field is str despite the dataclass type hints. 

These casts were to get things working, but yes - this is a bad solution and we should have better type enforcement here (and possibly repo-wide)

Based on what I can find, we can resolve this by casting the types at the object's init time for each field (kinda tedious) or by using something like pydantic (yet another package to manage). 

Should we attack this in another issue/PR or try to resolve it here?

api/api.py
vmpuri: See previous re: typecasting for dataclasses.

api/api.py
Jack-Khuu: Casting at init time makes sense, since downstream users shouldn't need to think about it

We can make it a separate PR or push the casting to init just for max_tokens and temperature

vmpuri: Created https://github.com/pytorch/torchchat/issues/1025 to track this, will merge this PR for now.

PR #1030, opened 2024-08-14
Add benchmarking scripts

Jack-Khuu:


scripts/benchmarking/benchmarking_linux.sh
```diff
@@ -0,0 +1,190 @@
+
```
byjlw, line 1: Add the license header at the top

scripts/benchmarking/benchmarking_mac.sh
```diff
@@ -0,0 +1,138 @@
+
```
byjlw, line 1: add license header at the top

scripts/benchmarking/benchmarking_linux.sh
byjlw, line 1: I'd drop benchmarking from the file names since they are already in a folder named benchmarking

byjlw (changes requested):

PR #1031, opened 2024-08-15
Rename use_tp to use_distributed

desertfire:
Summary: use_tp was renamed to use_distributed in https://github.com/pytorch/torchchat/pull/873, but the PR missed generate.py, which caused "AttributeError: 'BuilderArgs' object has no attribute 'use_tp'" when running generate with --profile.

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -160,9 +160,9 @@ class Generator:
         # global print
         #    from tp import maybe_init_dist
         #    rank = maybe_init_dist()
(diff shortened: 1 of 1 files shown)


Jack-Khuu (approved):

Jack-Khuu:
Failures seem mostly unrelated and should be fixed with the pin Bump

Worth a scroll to double check, but you should be good

PR #1037, opened 2024-08-16
style: add move model print

kuizhiqing:
After run
```
python3 torchchat.py download llama3.1
```
the last print would be
```
Converting meta-llama/Meta-Llama-3.1-8B-Instruct to torchchat format...██████████| 9.09M/9.09M [00:02<00:00, 3.43MB/s]
known configs: ['13B', '70B', 'CodeLlama-7b-Python-hf', '34B', 'Meta-Llama-3.1-8B', 'stories42M', '30B', 'stories110M', 'Meta-Llama-3.1-70B', '7B', 'stories15M', 'Mistral-7B', 'Meta-Llama-3-70B', 'Meta-Llama-3-8B']
Model config {'block_size': 2048, 'vocab_size': 128256, 'n_layers': 32, 'n_heads': 32, 'dim': 4096, 'hidden_dim': 14336, 'n_local_heads': 8, 'head_dim': 128, 'rope_base': 500000.0, 'norm_eps': 1e-05, 'multiple_of': 1024, 'ffn_dim_multiplier': 1.3, 'use_tiktoken': True, 'max_seq_length': 8192, 'use_scaled_rope': True}
Moving checkpoint to /Users/joeque/.torchchat/model-cache/downloads/meta-llama/Meta-Llama-3.1-8B-Instruct/model.pth.
Done.
```
from function `download_and_convert`, it was confusing that we cannot find any files in it.

I've found that the dir printed was a tmep dir and there is a move action in `download.py`,
So, add a print before move would be better.


Jack-Khuu (approved):
Thanks for the contribution!

PR #1039, opened 2024-08-16
[WIP][Draft only]. Integrate tune model arch, load weights via safetensors, add a lot of graph and model debug and tracing utils

lessw2020:
This PR is in progress but to checkpoint progress:
1 - Now uses torchtune model builders to create the model architecture (i.e. llama 3 8b).  Previously it used pure HF model definitions which was working to the point of running prefill.  
2 - Loads the model via meta init, traces the model with a fake tensor, creates the pipeline and segments the pipeline
3 - SPMD style, loads the stages weights from HF safetensors
4 - Creates real inputs
5 - Runs prefill schedule to generate first 'next' tokens

7 - Adds a number of utilities for safetensor loading, remapping, permuting and stage state_dict loading.  
8 - Adds a number of debug utilities for checking fx graph modules and models arch for dtypes, structure and faketensor/meta tensor inspection. 

## What works now:
All the model and tracing and weight loading is all working:

Screenshot 2024-08-15 at 4 30 40 PM

## What is not working:
Currently the model checkpoints are somehow also being interpreted as fake tensors so we end up with a model that is still all fake tensors, and thus on pipeline run, it states faketensor dispatch has no c10d::Send op. 

## What's next:
need to understand how even the checkpoints are coming in as fake tensors instead of real tensors...this feels like the fakemode context is somehow continuing throughout this loading process.  

## How do I run this? 
launch with bash meta_inference.sh
That will call into pp_meta.py which is the main processing file.
From there modeling utils and hf_utils apis are pulled in and used. 
There is a ton of debug info being logged atm, so  expect a lot of action in your console.


lessw2020:
closing as this has all been ported forward to the new built in native parallel approach.

PR #1055, opened 2024-08-23
Add max-autotune for CPU, update profile and fix next token calculation

yanbing-j:
This PR is to add max-autotune for CPU in torch.compile. Meanwhile, split first token and next token in the log print.

### DIFF
--- a/torchchat/generate.py
+++ b/torchchat/generate.py
@@ -115,6 +115,7 @@ class GeneratorArgs:
     compile_prefill: bool = False
     speculate_k: int = 5
     sequential_prefill: bool = False
+    max_autotune: bool = False
 
     def __post_init__(self):
         if self.compile_prefill and self.sequential_prefill:
@@ -159,6 +160,7 @@ class GeneratorArgs:
             compile_prefill=args.compile_prefill,
             speculate_k=args.speculate_k,
             sequential_prefill=sequential_prefill,
+            max_autotune=args.max_autotune,
         )
 
 
@@ -615,6 +617,7 @@ class Generator:
         tokens = self.tokenizer.encode(string)
         if bos:
             tokens = [self.tokenizer.bos_id()] + tokens
+        print("size after encode_tokens:", len(tokens))
         return torch.tensor(tokens, dtype=torch.int, device=device)
 
     def _callback(self, x, *, buffer, done_generating):
@@ -661,13 +664,18 @@ class Generator:
                     False  # Bug with cudagraph trees in this case
                 )
 
+            if self.builder_args.device == "cpu":
+                kwargs = {"mode": "max-autotune"} if generator_args.max_autotune else {}
+            else:
+                kwargs = {"mode": "reduce-overhead"}
+
             if self.is_speculative:
                 self.model_forward = torch.compile(
-                    self.model_forward, mode="reduce-overhead", fullgraph=True
+                    self.model_forward, fullgraph=True, **kwargs
                 )
 
             self.decode_one_token = torch.compile(
-                self.decode_one_token, mode="reduce-overhead", fullgraph=True
+                self.decode_one_token, fullgraph=True, **kwargs
             )
 
             if generator_args.compile_prefill:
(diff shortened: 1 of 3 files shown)


Jack-Khuu:
Thanks for the plumbing through max_autotune @yanbing-j that part looks great to me

@vmpuri Can you give the profiling/token calculation a quick pass though?

yanbing-j:
@Jack-Khuu @vmpuri Thanks for the review! Let me clarify something in updating profile and fix next token calculation.

In profiling, I add the logic of print profiling table both for CPU and GPU. In next token calculation, `t` includes first token (prefill) and next token (decode_n_tokens). `num_tokens_generated` is next token length, therefore, `t` and `num_tokens_generated` are not match. I suppose this should be a typo when adding first token time. And I also add first token latency and next token latency in the print log seperately.

yanbing-j:
@Jack-Khuu @vmpuri  Could you please help review and merge this PR?

Jack-Khuu (commented):
Thanks @yanbing-j for the changes and pinging again 

Some minor changes, then we're gtg

yanbing-j:
@Jack-Khuu  Thanks for the comments! Please review again!

Jack-Khuu (approved):
Thanks for updating the logging, everything looks good

Give the merge conflict (should be minor) a look and it's set

yanbing-j:
@Jack-Khuu  Thanks for the review!

I have rebased on main branch. And I also hide `tokens_sec` with jit compilation time. Then the average throughput will be more accurate.

yanbing-j:
Hi @Jack-Khuu , please help merge this PR. Thanks!

yanbing-j:
Hi @Jack-Khuu , please help review and merge this PR. Thanks!

Jack-Khuu:
Thanks for following up. I'm debugging some weird behavior with the output messages at the moment (on main)

Will merge this in once that's resolved

yanbing-j:
@Jack-Khuu Thanks! All the CI passes.

torchchat/generate.py
```diff
@@ -661,13 +664,22 @@ def chat(
...
-                    self.model_forward, mode="reduce-overhead", fullgraph=True
+                    self.model_forward, fullgraph=True, **kwargs
                 )
 
             self.decode_one_token = torch.compile(
-                self.decode_one_token, mode="reduce-overhead", fullgraph=True
+                self.decode_one_token, fullgraph=True, **kwargs
             )
 
             if generator_args.compile_prefill:
```
jgong5, line 685: Pass `kwargs` to compile_prefill model too?

Jack-Khuu:
Thanks again for the changes @yanbing-j 

Merging in (I'll tweak some nits in a separate PR)

sanchitintel:
@yanbing-j, with these changes, I observed a different behavior from [earlier](https://github.com/pytorch/pytorch/issues/134449#issuecomment-2318887128) while running `generate.py`.
Not sure if it's because of this PR, or because of the other changes introduced in `torchchat`.


With this PR's commits merged onto torchchat's main branch, I [see a lot of auto-tuning benchmarking results, even for the same shapes](https://gist.github.com/sanchitintel/d07678cd03548a3644db6e106d4b8e9b), after I run `python3 torchchat.py generate llama3.1 --prompt 'Hello my name is' --quantize '{"linear:int8": {"bitwidth": 8, "groupsize": 0}}' --compile --num-samples 5 --device cpu --tokenizer-path /localdisk/sanchitj/llama_3.1/original/tokenizer.model --max-autotune`

Is it expected behavior?

sanchitintel:
@yanbing-j, turns out `torch._inductor.config.trace.log_autotuning_results = True` is simply displaying more auto-tuning results, but I'm surprised that auto-tuning is being done for input shapes already encountered earlier (but this concern is orthogonal to this PR).

yanbing-j:
@sanchitintel The logs you observed from autotuning is printed by setting `torch._inductor.config.trace.log_autotuning_results = True`.

sanchitintel:
Thanks, @yanbing-j! That's what I meant. 

Should we disable it, as it's too verbose? Even without `torch._inductor.config.trace.log_autotuning_results = True`, we get all benchmarking logs for all unique input shapes. Thanks!

torchchat/generate.py
yanbing-j: Yes, miss this feature. Create another PR to support. https://github.com/pytorch/torchchat/pull/1112

yanbing-j:
@sanchitintel Remove this config in https://github.com/pytorch/torchchat/pull/1112.

PR #1059, opened 2024-08-24
Initial add of distributed model

kwen2501:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #1060
* __->__ #1059

Use `parallelize_module` in model.

Added files:

`model_dist.py`: a mirror of model.py with Tensor Parallelism baked in.
`dist_run.py`: toy example of how to run the model in distributed way.

Test:
`torchrun --nproc-per-node 2 dist_run.py`

PR #1060, opened 2024-08-24
Add pipeline parallel

kwen2501:
* __->__ #1060
* #1059

PP + TP now working.

### DIFF
--- a/build/model_dist.py
+++ b/build/model_dist.py
@@ -30,27 +30,35 @@ Rowwise = RowwiseParallel(use_local_output=False)
 device_mesh = None
 
 
-class Transformer(nn.Module):
-    def __init__(self, config: TransformerArgs) -> None:
+class TransformerStage(nn.Module):
+    def __init__(self, config: TransformerArgs, stage_idx: int, n_stages: int) -> None:
         super().__init__()
         self.config = config
+        self.stage_idx = stage_idx
+        self.n_stages = n_stages
+        self.layers_per_stage = config.n_layers // n_stages
 
         # Get device mesh
         global device_mesh
         if device_mesh is None:
             device_mesh = _mesh_resources.get_current_mesh()
 
-        tok_embeddings = nn.Embedding(config.vocab_size, config.dim)
-        self.tok_embeddings = parallelize_module(
-            tok_embeddings,
-            device_mesh,
-            RowwiseParallel(input_layouts=Replicate()),
-        )
-        self.layers = nn.ModuleList(
-            TransformerBlock(config) for _ in range(config.n_layers)
-        )
-        self.norm = RMSNorm(config.dim, eps=config.norm_eps)
-        self.output = nn.Linear(config.dim, config.vocab_size, bias=False)
+        if stage_idx == 0:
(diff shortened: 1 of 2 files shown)


build/model_dist.py
```diff
@@ -30,27 +30,35 @@
...
+            self.tok_embeddings = parallelize_module(
+                tok_embeddings,
+                device_mesh,
+                RowwiseParallel(input_layouts=Replicate()),
+            )
+
+        # Use ModuleDict so that each layer can be assigned its layer ID in the original model
+        self.layers = nn.ModuleDict()
+        for layer_id in range(self.layers_per_stage * stage_idx, self.layers_per_stage * (stage_idx + 1)):
+            self.layers[str(layer_id)] = TransformerBlock(config)
```
lessw2020, line 57: this is pretty clever!

build/model_dist.py
```diff
@@ -67,7 +75,7 @@ def setup_caches(self, max_batch_size, max_seq_length):
         max_seq_length = find_multiple(max_seq_length, 8)
         self.max_seq_length = max_seq_length
         self.max_batch_size = max_batch_size
-        for b in self.layers:
+        for b in self.layers.values():
```
lessw2020, line 78: block instead of b for clarity?

dist_run.py
```diff
@@ -4,32 +4,65 @@
...
-from build.model_dist import Transformer
+from build.model_dist import TransformerStage
 
 # Model config
 config = TransformerArgs.from_name("Transformer-2-7b-chat-hf")
 print(config)
 
 # Construct a device mesh with available devices (multi-host or single host)
-device_mesh = dist.init_device_mesh("cuda", (2,), mesh_dim_names=("tp",))
+device_mesh = dist.init_device_mesh("cuda", (2, 2), mesh_dim_names=("pp", "tp"))
```
lessw2020, line 22: if this file expands in future, maybe better to functionalize these different stages ala 
create_device_mesh(mesh_shape=(2,2))
setup_model
create_pipeline_stage
etc. 
It's clear atm what's happening but if this file becomes larger/more dynamic then maybe it would make sense in the future.

lessw2020 (approved):
looks great - the fancy work with module dict for assigning layers to stage is very clever.

build/model_dist.py
kwen2501: from original model.

dist_run.py
kwen2501: Sure makes sense.

dist_run.py
```diff
@@ -4,34 +4,67 @@
...
-    output = model(input_ids, input_pos)
+    mbs = 2                         # number of micro-batches
+    mb_size = 1                     # micro-batch size
+    batch_size = mbs * mb_size      # total batch size
+    seqlen = 4096                   # sequence length
+    dim = 4096                      # embedding dimension
+
+    # Example input for pipeline stages
+    mb_ids = torch.randint(0, config.vocab_size, (mb_size, seqlen), device=device)
+    activation = torch.rand(mb_size, seqlen, dim, device=device)
```
awgu, line 49: does `PipelineStage` requires that these are materialized on CUDA for them to be `input_args`?

PR #1070, opened 2024-08-28
Lowbit

metascroy:


metascroy:
@Jack-Khuu adding you here to gauge your thoughts on integration of the new torchao kernel's in torchchat's quantization API (also added you on the torchao diff D62394341).  Some clean up is still needed.

Overall, ARM CPU performance of the new kernels on llama3.1-8B is competitive across surfaces:

* 9.6 tokens/sec in torchchat generate.py eager (P1577139816)
* 17.9 tokens/sec in torchchat generate.py compile (P1577141735)
* 17.1 tokens/sec in torchchat generate.py AOTI (P1577169271)
* 18.6 tokens/sec in torchchat AOTI C++ runner (P1577229019)

In addition, ExecuTorch export works (P1577157356), but running the ExecuTorch *.pte file leads to missing op errors because we haven't written the ExecuTorch wrappers for the kernels yet (P1577172744).

One call out I want to make is the slow model loading perf in torchchat's generate.py.  The AOTI C++ runner loads the *.so and runs the model instantly and feels much, much smoother than the python version (which takes ~30sec to load the model).  We should fix this experience.

Jack-Khuu:
Thanks for the PR and tag. Numbers are looking great and the API you're showing is inline with how we've been leveraging AO. 

Like you mentioned in the Diff, it's worth connecting with the AO folk about the tensor subclass story

Jack-Khuu:
Let me know if you need help with refactoring or integrating this into torchchat

metascroy:
> Let me know if you need help with refactoring or integrating this into torchchat

I'll add you as a reviewer on the integration PRs.  Hopefully they will be ready soon.

.pins/torchao-pin.txt
```diff
@@ -0,0 +1 @@
+85d03de43160328eaf350e7ec3877d3d7b57da50
```
metascroy, line 1: TODO: update to commit hash after [D62394341](https://www.internalfb.com/diff/D62394341) lands.

scripts/install_utils.sh
```diff
@@ -124,3 +124,54 @@ install_executorch_libs() {
...
+
+clone_torchao() {
+  echo "Cloning torchao to ${TORCHCHAT_ROOT}/${TORCHAO_BUILD_DIR}/src"
+  rm -rf ${TORCHCHAT_ROOT}/${TORCHAO_BUILD_DIR}/src
+  mkdir -p ${TORCHCHAT_ROOT}/${TORCHAO_BUILD_DIR}/src
+  pushd ${TORCHCHAT_ROOT}/${TORCHAO_BUILD_DIR}/src
+  echo $pwd
+
+  cp -R /Users/scroy/fbsource/fbcode/pytorch/ao .
+  # git clone https://github.com/pytorch/ao.git
```
metascroy, line 136: TODO: uncomment after [D62394341](https://www.internalfb.com/diff/D62394341) lands and commit hash is updated.

generate.py
```diff
@@ -395,6 +395,13 @@ def decode_n_tokens(
                     )
                     input_pos += 1
                     break
+            if _i == 1:
```
metascroy, line 398: Remove this before final landing.  It was to do more accurate tokens/sec measurement during development, especially for torch.compile.

scripts/install_utils.sh
```diff
@@ -124,3 +124,54 @@ install_executorch_libs() {
 
   install_executorch_python_libs $1
 }
+
+clone_torchao() {
```
Jack-Khuu, line 128: We have a pin for 0.5.0 that you might be able to piggy back off of. Can save you some effort

https://github.com/pytorch/torchchat/pull/1136

Jack-Khuu (commented):
Overall I think your plan for getting the custom ops in and supporting the quants seems good to me

quantization/quantize.py
```diff
@@ -92,9 +92,21 @@ def quantize_model(
...
-                    groupsize=q_kwargs["groupsize"], device=device, precision=precision
-                )
+                if quantizer == "linear:a8wlow":
+                    quant_handler = ao_quantizer_class_dict[quantizer](
+                        device=device,
+                        precision=precision,
+                        bitwidth=q_kwargs.get("bitwidth", 4),
+                        groupsize=q_kwargs.get("groupsize", 128),
+                        has_weight_zeros=q_kwargs.get("has_weight_zeros", False),
+                        squeeze_unsqueeze_dim0=True,
```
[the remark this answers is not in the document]
metascroy: I modeled the args after Int8DynActInt4WeightQuantizer, and used "groupsize" instead of "group_size" to match the experience.

But Int8DynActInt4WeightQuantizer doesn't have bitwidth and has_weight_zeros as concepts, but happy to rename those to whatever you think best.

scripts/install_utils.sh
metascroy: This is cloning torchao from source, but the pin looks like it is from pip?

Jack-Khuu: We can decide on it when your diff lands

Big thing here is that AO doesn't have Mac nightlies so I'm fine with us moving back to direct clones for mac

PR #1096, opened 2024-08-31
[Consolidate] Move parallelization code into a .distribute() method

kwen2501:
The goal is to make the difference between `model.py` and `model_dist.py` as small as possible, and eventually consolidate them.

**[Consolidate TP]**

A first step is to move the parallelization code out from `.__init__()` into a separate `.distribute()` method, which is now an optional call:

```
model = Transformer(config)
model.distribute(tp_mesh)
```

`model.distribute` is a recursive call -- when hitting the child modules, it calls PyTorch's `parallelize_module` API with column-wise or row-wise styles. 

**[Consolidate PP]**

This is mostly cosmetic change -- naming `TransformerStage` back to `Transformer`, moving `stage_idx` into `config` etc.

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -19,8 +19,8 @@ from distributed.safetensor_utils import (
 )
 from distributed.utils import Color as color
 from torch.distributed.pipelining import PipelineStage, ScheduleGPipe
-from torchchat.model import ModelArgs, TransformerArgs
-from torchchat.model_dist import TransformerStage
+from torchchat.model import ModelArgs
+from torchchat.model_dist import Transformer
 from torchchat.utils.build_utils import get_precision
 
 MODEL_NAME = "Transformer-2-7b-chat-hf"
@@ -84,8 +84,12 @@ def main():
     nstages = pp_mesh.size()
     device = torch.device(f"cuda:{rank}")
 
+    # Fill in PP configs
+    config.stage_idx = pp_rank
+    config.n_stages = nstages
+
     with device:
-        model = TransformerStage(config, pp_rank, nstages)
+        model = Transformer(config)
 
     model.setup_caches(1, 4096)
     # TODO: refine this .to once we start using fp8 for KV cache
--- a/dist_run.py
+++ b/dist_run.py
@@ -85,11 +85,14 @@ def main():
     device = torch.device(f"cuda:{rank}")
 
     with device:
-        with tp_mesh:
-            model = TransformerStage(config, pp_rank, nstages)
-            model.setup_caches(1, 4096)
-            # TODO: refine this .to once we start using fp8 for KV cache
-            model = model.to(model_dtype)
+        model = TransformerStage(config, pp_rank, nstages)
+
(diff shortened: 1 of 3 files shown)


torchchat/model_dist.py
```diff
@@ -107,22 +88,38 @@ def setup_caches(self, max_batch_size, max_seq_length):
...
         assert self.freqs_cis is not None, "Caches must be initialized first"
         if input_pos is None:
             input_pos = torch.arange(x.shape[1], device=x.device, dtype=torch.long)
         mask = self.causal_mask[None, None, input_pos]
         freqs_cis = self.freqs_cis[input_pos]
 
-        if self.stage_idx == 0:
-            x: DTensor = self.tok_embeddings(x)
-            # TODO: sequence parallelize this
+        if hasattr(self, "tok_embeddings"):
```
lessw2020, line 113: minor - is it possible to just check stage_idx as before since embeddings will always be on stage0, or alternatively do this hasattr in a post init type of deal and set a self.hasnorm bool property so we're not having to run through all the class dicts * every hasattr check, over and over on every token coming through.
My (minor) concern is that having all these attr checks in a forward could add up to a bit of perf hit in the larger models vs a bool or int check like before.  
More important to ensure things work first so this is fine but maybe just flag a TODO for perf streamlining.

torchchat/model_dist.py
```diff
@@ -107,22 +88,38 @@ def setup_caches(self, max_batch_size, max_seq_length):
...
+            x = self.tok_embeddings(x)
 
         for _, layer in self.layers.items():
             x = layer(x, input_pos, freqs_cis, mask)
 
-        if self.stage_idx == self.n_stages - 1:
+        if hasattr(self, "norm"):
```
lessw2020, line 119: same as above - wondering if these can be n_stages-1 as before for faster check.

lessw2020 (approved):
lgtm!  Thanks for the quick consolidation work!

dist_run.py
```diff
@@ -84,12 +84,19 @@ def main():
...
     with device:
-        with tp_mesh:
-            model = TransformerStage(config, pp_rank, nstages)
-            model.setup_caches(1, 4096)
-            # TODO: refine this .to once we start using fp8 for KV cache
-            model = model.to(model_dtype)
+        model = Transformer(config)
+
+    model.setup_caches(1, 4096)
+    # TODO: refine this .to once we start using fp8 for KV cache
```
mikekgfb, line 95: If KV cache is the only consideration in type conversion, you might just invert the order?

torchchat/model_dist.py
kwen2501: Makes sense. I am changing it to `if self.tok_embeddings:`. Correspondingly, at init, `None` will be assigned if not stage 0.

kwen2501:
`test-gguf-util (macos-14)` fails in trunk too, so it is unrelated to this PR.
Merging.

PR #1097, opened 2024-09-01
safetensor updates: add attn head permute, update gating layer mapping

lessw2020:
This PR:
1 - updates the q_proj and k_proj permuting to permute based on attn head dimensions (in line with torchchat main)
2 - updates the gating layer mapping to be w3 as well as remap w1 and w2 to correct up and down projection.

With these changes, we have full verification of 100% weight matching of all sampled weights when comparing stage weights to their subsection of single gpu weights ala:
Screenshot 2024-08-31 at 8 03 52 PM

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -8,7 +8,7 @@
 # torchrun --nproc-per-node 4 dist_run.py
 import torch
 import torch.distributed as dist
-from distributed.dtensor_utils import find_cpu_tensors, record_module_dtypes
+from distributed.dtensor_utils import find_cpu_tensors
 from distributed.logging_utils import setup_logging
 
 # TODO - these are not distributed specific, consider moving to new package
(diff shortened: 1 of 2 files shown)


dist_run.py
```diff
@@ -41,10 +40,17 @@ def _create_device_mesh(mesh_dimensions):
     return dist.init_device_mesh("cuda", mesh_dimensions, mesh_dim_names=("pp", "tp"))
 
 
-def _load_model_weights(stage_module, hf_model_name, device, logger):
+def _load_model_weights(stage_module, hf_model_name, device, logger, model_config):
```
kwen2501, line 43: nit: remove logger from function signature?
nit: comment on what `model_config` is purposed for?

kwen2501 (approved):
Thanks for the digging! Looks good to me!

dist_run.py
lessw2020: resolved both!

PR #1098, opened 2024-09-02
[distributed] add verification_utils.py - collection of validation utils for verifying various model aspects

lessw2020:
This PR:
1 - adds verification_utils.py, which holds a collection of different functions useful for various verification tasks.  Most of these are primarily for checking things during dev work and not so much during actual runtime.  
The exception is the find_cpu_tensors is currently run as a check after all stages have been loaded during dist inference.

Summary of functions:

**record_module_dtypes(module):** Record the dtypes of all parameters and buffers in a module and return a dictionary of dtype -> list of names

**find_cpu_tensors(module):** Find all CPU tensors in a module and return a list of their names

**inspect_module_tensors**  Inspect an nn.Module and return information about its tensor types.

**torch_in_fake_mode() -> bool:**   Check if torch is in fake mode.

**get_tensor_type(tensor: torch.Tensor)**  Get the type of a tensor (regular, fake, or meta).

**extract_and_save_weights(model, output_file):** Extract the first 4 weights from each module (including nested ones) and nested buffers in a model and save them to a CSV file.
    Supports regular tensors, DTensors, and nested structures.

**compare_weight_files(file1, file2, tolerance=1e-6):** Compare two CSV files containing model weights and report any differences.

**enumerate_model_details(model, prefix="", output_file=None):**  Prints information about the model's modules and parameters.


kwen2501:
Agree with the move.

Otherwise, would be great to clean up the code in future (when development settles down). In particular, we can think of the following aspect --
when an end user install this library, would they need these codes?

kwen2501 (approved):
Stamping to unblock.
Mostly movement of debug utilities.

PR #1099, opened 2024-09-03
[Distributed] Enable sequence parallel

kwen2501:
Sequence parallel = Tensor parallel + dividing sequence by `tp_degree` at layer boundary

In addition to the op parallelism achieved by TP, SP reduces the activation size transferred between pipeline stages by `tp_degree`X.

The PR includes 4 parts:
1. Give SP styles to `parallelize_module` calls (this part mainly mimics what's used in torchtitan).
2. Before the attention layer, gather the sequence back.
3. Adjust example input to pipeline stages (divided by `tp_degree`).
4. Adjust the code for creating `input_pos` -- we still need to use the full seq length instead of `x.shape[1]` to create it bc `x` could have been cut in SP cases. 

Test:
`torchrun --nproc-per-node 8 dist_run.py`
(PP=2, TP=4)

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -75,18 +75,30 @@ def main():
         raise ValueError(f"Config file not found for model id {hf_model_name}")
     logger.info(f"Using HF model weights from {hf_model_name}")
 
-    mesh_dimensions = (2, 2)
-    device_mesh = _create_device_mesh(mesh_dimensions)
+    # Assuming 2 pipeline stages, feel free to change this as long as the
+    # asserts are satisfied
+    pp_degree = 2
+    assert world_size % pp_degree == 0
+    assert config.n_layers % pp_degree == 0
 
+    # Sequence parallel is enabled in this program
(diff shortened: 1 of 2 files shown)


dist_run.py
```diff
@@ -97,16 +109,18 @@ def main():
 
     # Distribute model on TP mesh
     model.distribute(tp_mesh)
-    logger.info(f"Model: {model}")
+    if rank == 0:
```
lessw2020, line 112: minor note - I enabled log filtering by rank in the run_dist.sh launcher, thus having rank specific filtering in the code might end up being confusing. (i.e. if debugging last pp stage via global filter and then not seeing expected logging b/c we have hardcoded filtering in the code). 
~~~
LOG_RANK=${LOG_RANK:-0,1,2,3}
~~~

lessw2020 (approved):
nice job!  this really completes all the main pieces (pp/tp/sp) for dist foundation.

dist_run.py
kwen2501: Ah, okay, reverting it

PR #1100, opened 2024-09-03
Switch gguf-util test to a smaller model to prevent timeout

Jack-Khuu:


.github/workflows/pull.yml
```diff
@@ -721,14 +721,14 @@ jobs:
       - name: Download GGUF files
         run: |
           mkdir gguf_files
-          wget -O gguf_files/llama-2-7b.Q4_0.gguf "https://huggingface.co/TheBloke/Llama-2-7B-GGUF/resolve/main/llama-2-7b.Q4_0.gguf?download=true"
-          ./llama.cpp/llama-quantize --allow-requantize gguf_files/llama-2-7b.Q4_0.gguf gguf_files/llama-2-7b.Q4_0.requant_F32.gguf F32
+          wget -O gguf_files/TinyLlama-1.1B-openorca.Q4_0.gguf "https://huggingface.co/TheBloke/TinyLlama-1.1B-1T-OpenOrca-GGUF/resolve/main/tinyllama-1.1b-1t-openorca.Q4_0.gguf?download=true"
```
Gasoonjia, line 724: Is not using original llama a good choice to verify torchchat correctness?

byjlw: We need a smaller model for our automatic build checks. Otherwise it takes to long and hangs. 
We should use periodic tests for the bigger models.

byjlw (approved):

.github/workflows/pull.yml
Jack-Khuu: Eventually we will want to, but this test doesn't feel like the right medium cadence for it, since
* We don't advertise GGUF support
* The related lines of code are low traffic, but the test slows down other developers 
* Larger model tests would be better suited for periodic daily tests

PR #1102, opened 2024-09-03
Merge model_dist into model

kwen2501:
Minimized the difference between model_dist and model, and merge the former into the latter.

What's new to model.py:

1. Tensor parallelism:
Added a `.distribute()` method to module classes which distributes the module in SPMD manner upon call.
Default: original model

2. Pipeline parallelism: 
`Transformer` now builds stage modules based on a `stage_idx`, including or excluding the `tok_embedding`, `output` linear, etc, and include the `TransformerBlock`s corresponding to current stage.
Default: original model

Top level user experience:
```
model = Transformer(config)
model.distribute(tp_mesh)
```


torchchat/model.py
```diff
@@ -210,18 +218,36 @@ class Transformer(nn.Module):
...
+
+        # Use ModuleDict so that each layer can be assigned its layer ID in the original model
+        self.layers = nn.ModuleDict()
+
+        for layer_id in range(
+            layers_per_stage * config.stage_idx, layers_per_stage * (config.stage_idx + 1)
+        ):
+            self.layers[str(layer_id)] = TransformerBlock(config)
+
+        self.norm = (
```
lessw2020, line 236: minor style nit - I feel like this is more readable expressed as a default and if true modify block ala:
~~~
# only last stage handles final norm and output:
self.norm = None
self.output = None
if config.stage_idx == config.n_stages-1:
     self.norm = RMSNorm()
     self.output = nn.linear()
~~~
rather than each having identical ternary if statement at the end of the init call.

mikekgfb:
This seems like goodness.  Looking at the code, I can't imagine performance degradation (with that caveat that more imports does increase startup time), but this might be one of these places where running some benchmark(s) before and after would make sense.  

Integrating distributed into model.py is a great step towards:
1 - quantize (f8, f4, i8, i4 i8)  and distribute: to run 405B on a single node?
2 - build a `class Generator` that runs distributed code, at a minimum for the API service, so that users can interact with large distributed models like any other torchchat model.

torchchat/model.py
```diff
@@ -343,13 +413,26 @@ def _unfuse_wqkv_state_dict(
...
+
     def forward(
         self,
         x: Tensor,
         freqs_cis: Tensor,
         mask: Tensor,
         input_pos: Optional[Tensor] = None,
     ) -> Tensor:
+        # Gather sequence back in case of sequence parallelism before attention
+        if isinstance(x, DTensor):
```
lessw2020, line 433: minor perf q:
could Transformer provide:
self.seq_parallel_degree
to TransformerBlock.Attention during init 
so that this isinstance(x) check doesn't have to be done on every forward?
Instead it would just be:
if self.seq_parallel > 1:

lessw2020 (approved):
looks great to me. 
very cool to see distributed fully baked in/native now.

torchchat/model.py
kwen2501: Good question on perf. 
On the other hand, seq_parallel > 1 does not necessarily mean x is a `DTensor` (which depends on whether `to_local` has been called), so calling `.redistributed` on it is an uncertain behavior.

torchchat/model.py
kwen2501: Makes sense. Modified.

torchchat/model.py
lessw2020: good point that x may not always be dtensor. 
Just for kicks, made a quick script to compare perf - it's not that big (~5%):
~~~
Rank 0:
Time taken for isinstance check: 0.004911 seconds
Time taken for int setting check: 0.004534 seconds
isinstance is 1.083x slower than int setting check
Rank 1:
Time taken for isinstance check: 0.004674 seconds
Time taken for int setting check: 0.004502 seconds
isinstance is 1.038x slower than int setting check
~~~

PR #1104, opened 2024-09-04
[distributed] ensure precise rope embeddings by creating frequencies in bf16

lessw2020:
This PR:
1 - addresses and fixes a mismatch that was present in distributed model vs single gpu where the rope embeddings were not an exact match.  
example:
~~~

5.4030e-01, 8.4147e-01.  (distributed)
Vs
5.3906e-01,  8.3984e-01. (single gpu rope)

6.4791e-01, 7.6172e-01
Vs
6.4844e-01,  7.6172e-01

7.3176e-01, 6.8156e-01
Vs
7.3047e-01,  6.7969e-01
~~~

The issue was we were creating the frequencies float32, and then later setting the model (llama2) to float16 to match the checkpoint dtype. 

The fix is to use torchchat set_precision and set that to torch.bfloat16 so that all cache creation happens in torch.bfloat16 and thus expected rope embeddings precision.

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -21,6 +21,7 @@ from distributed.safetensor_utils import (
 from distributed.utils import Color as color
 from torch.distributed.pipelining import PipelineStage, ScheduleGPipe
 from torchchat.model import ModelArgs, Transformer
+from torchchat.utils.build_utils import set_precision
 
 logger = setup_logging(__name__)
 
@@ -28,6 +29,7 @@ MODEL_NAME = "Transformer-2-7b-chat-hf"
 NAME_TO_HF_MODEL_ID_AND_DTYPE = {
     "Transformer-2-7b-chat-hf": ("meta-llama/Llama-2-7b-chat-hf", torch.float16),
 }
+CACHE_PRECISION = torch.bfloat16
 
 
 def _init_distributed():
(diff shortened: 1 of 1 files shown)


dist_run.py
```diff
@@ -114,8 +119,6 @@ def main():
         model = Transformer(config)
 
     model.setup_caches(1, 4096)
-    # TODO: refine this .to once we start using fp8 for KV cache
-    model = model.to(model_dtype)
```
kwen2501, line 118: Curious -- 
If set/get precision are only effective for KV cache, why removing this `model.to` line is okay?

kwen2501 (approved):
Had one question. Otherwise, looks good to me.

dist_run.py
kwen2501: Or, are you saying that the purpose of this line was previously to cast the dtype of the cache **only**?

lessw2020: correct - I was only using this line to handle the dtype of the caches.  The rest of the weights are all being set in the checkpoint loading.
I also want to note that this set_precision is not going to be a long term fix b/c it's setting *two* caches dtype - right now it sets both kvcache and rope cache, but longer term we will want fp8/fp6/fp4 kvcache and yet continue with bf16 rope.. so this is not a perm fix, but resolves things for now.

PR #1108, opened 2024-09-05
Minor Readme changes

byjlw:
get rid of comments in the install command so people can copy paste the entire thing.

be explicit in the directions that the runner is c++


ali-khosh (approved):
Reviewed changes to the readme file to clarify what the runner does.

PR #1110, opened 2024-09-05
[distributed] integrate chat tokenizers, and add llama3-8B model option

lessw2020:
This PR:
1 - integrates the chat Tokenizers by using the TokenizerArgs and _inititialize_tokenizer functions from builder.py. 
With the _build_chat_tokenizer() you can instantiate the same tokenizers as installed by chat (rather than using HF tokenizer). 

example:
~~~
[rank0]:2024-09-05:15:57:26,835 INFO     [dist_run.py:83] using tokenizer = tokenizer.tiktoken.Tokenizer
~~~
and
~~~
[rank0]:2024-09-05:16:01:01,716 INFO     [dist_run.py:83] using tokenizer = sentencepiece.SentencePieceProcessor
~~~

2 - adds llama3-8B instructional as a valid model for dist.

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -4,30 +4,44 @@
 # This source code is licensed under the license found in the
 # LICENSE file in the root directory of this source tree.
 
+import os
+from pathlib import Path
+from types import SimpleNamespace
+from typing import Any, Dict
+
 # Run command:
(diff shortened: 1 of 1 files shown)


kwen2501 (approved):
lgtm

PR #1111, opened 2024-09-06
Torchao install from pypi instead of github

Gasoonjia:
Met some issues when installing torchao on gpu server.
This pr uses nightly pip installation on torchao to replace the original GitHub installation to solve the installation issue. Also ping the torchao to specific nightly version to avoid macos build issue.


Jack-Khuu:
Abandoning outdated diff

PR #1112, opened 2024-09-06
Add max-autotune option for prefill

yanbing-j:
This PR is to add max-autotune option for prefill, and remove autotuning log config. Users can autotuning log config by themselves.


yanbing-j:
@jgong5 @sanchitintel Could you please help review this PR?

Jack-Khuu (approved):
Thanks for the followup

PR #1115, opened 2024-09-06
Bumped up core pytorch version

Gasoonjia:
ET has updated their core pytorch version into 0901. TC also need to support that


Jack-Khuu:
can you check on the runner-et test, seems legit

malfet (approved):

Jack-Khuu:
closing outdated pr

PR #1116, opened 2024-09-08
Fix benchmarking statistics computation when profiler is used in generate.py

sanchitintel:
Currently, in `torchchat/generate.py`, benchmarking time includes time for printing the table & saving Chrome-viewable trace, so incorrect performance data for `Total throughput` and `Next token throughput` is being printed when the PyTorch profiler is enabled. The difference is significant (only verified on CPU. With int8 weight-only quantization, it resulted in a difference of 9 tokens/second, perhaps because I had also enabled shapes & memory profiling, which would've increased the size of the trace).

For verification, manually enabled profiling in `generate.py` because using `--profile` argument with `torchchat.py` did not work with HEAD commit 64510a33cd09b67b411e939ae0953d403724113c. I'm guessing its support is WIP.

### DIFF
--- a/torchchat/generate.py
+++ b/torchchat/generate.py
@@ -834,6 +834,8 @@ class Generator:
                 generator_args.compile or generator_args.compile_prefill
             )
             compilation_time = time.perf_counter() - t0
+            device_sync(device=self.builder_args.device)
+            t = time.perf_counter() - t0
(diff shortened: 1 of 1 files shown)


sanchitintel:
Hi @Jack-Khuu, can you please review this PR? Thanks!

Jack-Khuu (approved):

PR #1120, opened 2024-09-09
[distributed] add stage metrics - total params per stage, total size and present it in a nicely formatted manner

lessw2020:
This PR:
1 - computes the total params per stage and total byte size, and presents that in a nicely formatted manner:

Screenshot 2024-09-08 at 5 46 00 PM

There are four underlying functions supporting this - one each for getting the relevant data, and then one each for formatting it nicely into a nice human readable format. 

The result is we can quickly monitor / verify that stages are relatively balanced in params / size.

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -23,7 +23,12 @@ from distributed.safetensor_utils import (
     get_hf_weight_map_and_path,
     load_safetensor_weights,
 )
-from distributed.utils import Color as color, get_module_size, get_num_params, bytes_to_readable
+from distributed.utils import (
+    Color as color,
+    get_module_size,
+    get_num_params,
+    bytes_to_readable,
+)
 from distributed.verification_utils import find_cpu_tensors
 from torchchat.cli.builder import TokenizerArgs, _initialize_tokenizer
 from torchchat.model import ModelArgs, Transformer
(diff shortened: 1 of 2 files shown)


dist_run.py
```diff
@@ -184,6 +187,13 @@ def main():
     logger.info(f"Loading weights for {pp_rank=} on {device=}")
     _load_model_weights(model, hf_model_name, device=device, model_config=config)
 
+    # info on stage size and params
+    stage_size, stage_size_formatted = get_stage_size(model)
```
kwen2501, line 191: nit:
can we make the API return just a number? At the place of logging, we can then use `bytes_to_readable` to get a string.

distributed/utils.py
```diff
@@ -49,7 +52,47 @@ def get_num_params(model: torch.nn.Module, exclude_embedding: bool = False) -> i
     num_params = sum(p.numel() for p in model.parameters())
     if exclude_embedding:
         num_params -= model.tok_embeddings.weight.numel()
-    return num_params
+    readable_num_params = format_model_params(num_params)
+    return readable_num_params
+
+
+def get_stage_size(stage):
```
kwen2501, line 59: nit: can we name it "get_module_size"? It seems to me that the function can be generally applied to `nn.Module` than just a stage. Also, we can type the input arg.

kwen2501 (approved):
Mostly LGTM. Some minor comments re API design.

dist_run.py
lessw2020: updated

distributed/utils.py
lessw2020: added typiing and changed name.

PR #1121, opened 2024-09-09
[distributed] add TrackTime, CUDATrackTime to monitor perf for weight loading per stage and future perf measurements

lessw2020:
This PR:
1 - adds a TrackTimer class to wrap perf_counter and make it easy to do perf timings.
2 - implements it to track weight loading time per stage:

Screenshot 2024-09-08 at 6 22 53 PM

Note - there is a similar timer class called measure_time in chat, but I added some additional rounding options,  unit options and also didn't want to use print for the logging.

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -112,8 +112,11 @@ class TrackTime:
 
 
 class CUDATrackTime:
-    """integrated class for perf timing via cuda events.
-    Note - this uses a default stream synchronize, and defaults to current device."""
+    """
+    Integrated class for perf timing via cuda events.
+    Note - this uses the default stream to synchronize, and defaults to current device.
+    The event.record() will create a context on the CUDA device matching the device used at init.
+    """
 
     def __init__(self, device=None, use_ms: bool = False, round_to: Optional[int] = 4):
         if device is None:
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -112,11 +112,23 @@ class TrackTime:
 
 
 class CUDATrackTime:
-    """integrated class for perf timing via cuda events"""
+    """integrated class for perf timing via cuda events.
+    Note - this uses a default stream synchronize, and defaults to current device."""
 
-    def __init__(self, use_ms: bool = False, round_to: Optional[int] = 4):
-        self.start_event = torch.cuda.Event(enable_timing=True)
-        self.end_event = torch.cuda.Event(enable_timing=True)
+    def __init__(self, device=None, use_ms: bool = False, round_to: Optional[int] = 4):
+        if device is None:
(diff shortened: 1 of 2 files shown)


kwen2501:
Does our weight loading include any CUDA ops? 
It seems this timer captures time on the CPU side.

lessw2020:
> Does our weight loading include any CUDA ops? It seems this timer captures time on the CPU side.

there's no express cuda ops beyond a .to from cpu to gpu, but that is a sync event.

That said, I agree we want to have cuda events for timing most everything else, so adding a cuda even timer class.  
I guess I'll do a sep PR for that.

lessw2020:
nvm, makes sense to just add it here.
Added a CUDATrackTime, exact same interface so either drops in with same usage pattern, but CUDATrackTime uses cuda events for potentially more precise timing for any underlying cuda kernel calls.

distributed/utils.py
```diff
@@ -76,3 +81,80 @@ class NoColor:
...
+        if self.active:
+            raise RuntimeError("Timer is already running. Use .stop() to stop it")
+        self.start_event.record()
+        self.active = True
+
+    def stop(self):
+        if not self.active:
+            raise RuntimeError("Timer is not running. Use .start() to start it")
+        self.end_event.record()
+        self.active = False
```
kwen2501, line 136: nit: can you add a note here that these methods (containing `.record()`) can create CUDA context on `current_device()` (whatever it is) if there hasn't been one? Thus for safety, `torch.cuda.set_device()` is strongly needed for `CUDATrackTime`.

distributed/utils.py
```diff
@@ -76,3 +81,80 @@ class NoColor:
...
+
+    def get_time(self):
+        if self.active:
+            raise RuntimeError("Timer is still running. Use .stop() to stop it")
+
+        torch.cuda.synchronize()  # Wait for all GPU operations to finish
```
kwen2501, line 142: nit: do you mind adding a note to `CUDATrackTime` that this timer is going to trigger device synchronization? Thanks!

kwen2501 (approved):
LGTM. Suggested some notes as the utils themselves make some CUDA calls (which can lead to unexpected effects if users are not aware)

distributed/utils.py
lessw2020: I've updated this to take a device and then to use that for setting the events.  
Also added comment re: creating context on the designated device.

distributed/utils.py
lessw2020: added comment in doc string re: synchronization.

kwen2501:
The new changes look good! Thanks!

kwen2501 (approved):

PR #1124, opened 2024-09-10
[distributed] add SingletonLogger to avoid duplicate logging, with custom formatting options including file name: line_number

lessw2020:
This PR:
1 - adds a new SingletonLogger to eliminate an issue that was popping up with duplicate logging. 

Before:
Screenshot 2024-09-09 at 2 28 43 PM

After:

Screenshot 2024-09-09 at 7 27 44 PM

2 - This logger adds a couple useful formatting improvements:

    a - option to not display the year in the timestamp to save space
    b - avoid putting INFO / DEBUG for the level, again to save space (WARNING / ERROR / CRITICAL) are always shown
    c - shows the calling filename and line number even though this is now a singleton setup
    d - strips ".py" extension so that it's filename:line_number and not filename.py:line_number, again to save space.

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -16,10 +16,13 @@ import torch.distributed as dist
 from torch.distributed.pipelining import PipelineStage, ScheduleGPipe
 
 from distributed.logging_utils import SingletonLogger
+
 # TODO - these are not distributed specific, consider moving to new package
-from distributed.safetensor_utils import (get_hf_config_file,
-                                          get_hf_weight_map_and_path,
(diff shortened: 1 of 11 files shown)


kwen2501:
What is the key line that makes the log singleton?

kwen2501:
I see. Thanks. I can see this being useful in multi-thread scenario.
Still have to think how it helped in multi-process scenario.
But, nice work. (We will probably run in multi-thread mode later)

PR #1126, opened 2024-09-10
CLI + Openai api multimodal support

vmpuri:
Components for the OpenAI API and torchchat CLI to support multimodal models.


Jack-Khuu (commented):
Still reviewing, intial items

PR #1128, opened 2024-09-10
Add missing Block size + Update Configs to not hardcode rope_scaling

Jack-Khuu:
1) Previously the block_size was not being included in the model_params.json of llama3/3.1 models so it just used a hard coded 2048. This PR adds them into the config.

2) For rope_scaling (llama 3.1), we were hardcoding the parameters inside of `apply_scaling()` instead of taking them in as TransformerArgs built from model_params. These happen to line up for 3.1, but this PR makes them an explicit read

### DIFF
--- a/torchchat/model.py
+++ b/torchchat/model.py
@@ -682,7 +682,9 @@ class RMSNorm(nn.Module):
 
 def apply_scaling(freqs: torch.Tensor, rope_scaling: Dict[str, Any]):
     # Check for the presence of the required keys
-    assert set(rope_scaling.keys()) >= {"factor", "low_freq_factor", "high_freq_factor", "original_max_position_embeddings"}
+    required_keys = {"factor", "low_freq_factor", "high_freq_factor", "original_max_position_embeddings"}
+    if not required_keys.issubset(rope_scaling.keys()):
+        raise ValueError(f"Missing required keys in apply_scaling. Expected: {required_keys}")
 
     scale_factor = rope_scaling["factor"]
     low_freq_factor = rope_scaling["low_freq_factor"]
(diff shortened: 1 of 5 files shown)


torchchat/model.py
```diff
@@ -503,12 +501,14 @@ def forward(self, x: Tensor) -> Tensor:
...
 
-def apply_scaling(freqs: torch.Tensor):
-    # Values obtained from grid search
-    scale_factor = 8
-    low_freq_factor = 1
-    high_freq_factor = 4
-    old_context_len = 8192  # original llama3 length
+def apply_scaling(freqs: torch.Tensor, rope_scaling: Dict[str, Any]):
+    # Check for the presence of the required keys
+    assert set(rope_scaling.keys()) >= {"factor", "low_freq_factor", "high_freq_factor", "original_max_position_embeddings"}
```
lessw2020, line 506: could be clearer with issubset and a raiseValueError for more informative error vs assert?
~~~
required_keys = {"factor", "low_freq_factor", "high_freq_factor", "original_max_position_embeddings"}
    if not required_keys.issubset(rope_scaling.keys()):
        raise ValueError(f"Missing required keys in apply_scaling. Expected: {required_keys}")
 ~~~

lessw2020 (approved):
lgtm! 
1 - I know I wasn't asked to review but since we did a lot of work getting rope embeddings happy for distributed wanted to check this. 
2 - verified no issue with distributed run
3 - left minor option for using issubset to potentially read a bit more clearly.

Jack-Khuu:
Thanks for the review, always welcome!! Good idea with the Error raise

PR #1130, opened 2024-09-11
Install torchtune on linux

Gasoonjia:
This PR tries to add torchtune support on linux, to have a unify specific version for us to test torchtune with torchchat on linux machine.

### DIFF
--- a/install/install_requirements.sh
+++ b/install/install_requirements.sh
@@ -95,6 +95,11 @@ LINUX_REQUIREMENTS_TO_INSTALL=(
 )
 
 PLATFORM=$(uname -s)
+
+# Install torchtune and torchao requirements for Linux systems using nightly.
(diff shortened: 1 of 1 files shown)


install/install_requirements.sh
```diff
@@ -41,13 +41,23 @@ fi
...
-# NIGHTLY_VERSION, you should re-run this script to install the necessary
+# PYTORCH_NIGHTLY_VERSION, you should re-run this script to install the necessary
 # package versions.
-NIGHTLY_VERSION=dev20240814
+PYTORCH_NIGHTLY_VERSION=dev20240814
+
+# Nightly version for torchvision
+VISION_NIGHTLY_VERSION=dev20240814
+
+# Nightly version for torchao
```
Jack-Khuu, line 55: Let's add another comment here that Mac is using a different install/version

Editors don't further unless the code fails

Jack-Khuu (approved):

PR #1133, opened 2024-09-11
[distributed] prefill (single and multi-prompt) and single prompt generation and decoding 

lessw2020:
This PR:
1 - enables single or batch prompt prefill and decoding. 
2 - adds three supporting functions for this (encode string, create padded prompt, batch decode next tokens). 
Single:
Screenshot 2024-09-11 at 2 41 11 PM

Batch - prompts:
Screenshot 2024-09-11 at 6 17 45 PM

and batch decode:
Screenshot 2024-09-11 at 6 17 07 PM

3 - enables single prompt decoding up to max_tokens with formatted output:

what_is-Pytorch_example


dist_run.py
```diff
@@ -122,6 +112,88 @@ def _load_model_weights(stage_module, hf_model_name, device, model_config):
         raise ValueError(f"Missing {num_missing_weights} weights")
 
 
+def _encode_strings(
+    strings: List[str],
+    tokenizer,
+    bos: bool = True,
+    device: str = "cuda",
```
kwen2501, line 119: nit: `device: torch.device`

dist_run.py
```diff
@@ -122,6 +112,88 @@ def _load_model_weights(stage_module, hf_model_name, device, model_config):
...
+        encoded_list.append(torch.tensor(tokens, dtype=dtype, device=device))
+    return encoded_list
+
+
+def _create_padded_prompts(
+    input_ids_list: List[torch.Tensor],
+    tokenizer,
+    seqlen: int,
+    start_pos: int,
+    device: str,
```
kwen2501, line 137: same

dist_run.py
```diff
@@ -122,6 +112,88 @@ def _load_model_weights(stage_module, hf_model_name, device, model_config):
...
+    Returns:
+        Tuple[torch.Tensor, List[int]]: A tuple containing the padded tensor and a list of prompt lengths.
+    """
+    pad_token_id = pad_token_id if pad_token_id is not None else tokenizer.eos_id()
+
+    # Find the maximum prompt length
+    max_prompt_len = max(ids.size(0) for ids in input_ids_list)
+
+    # Calculate the buffer size
+    max_new_tokens = max(0, min(seqlen - start_pos, seqlen - max_prompt_len))
```
kwen2501, line 152: nit: do you mind sharing a comment on what `seqlen - start_pos` stand for?

dist_run.py
```diff
@@ -122,6 +112,88 @@ def _load_model_weights(stage_module, hf_model_name, device, model_config):
         raise ValueError(f"Missing {num_missing_weights} weights")
 
 
+def _encode_strings(
+    strings: List[str],
+    tokenizer,
+    bos: bool = True,
```
kwen2501, line 118: nit: what does `bos` mean?

run_dist.sh
```diff
@@ -1,4 +1,4 @@
-export CUDA_VISIBLE_DEVICES=4,5,6,7
+#export CUDA_VISIBLE_DEVICES=4,5,6,7
 PORT=${1:-29501}
 NGPU=${NGPU:-"4"}
 LOG_RANK=${LOG_RANK:-0,1,2,3}
```
kwen2501, line 4: nit: maybe we should remove this script and just keep an example in a document?

dist_run.py
```diff
@@ -167,6 +258,17 @@ def main():
     pp_mesh = device_mesh["pp"]
     tp_rank = tp_mesh.get_local_rank()
     pp_rank = pp_mesh.get_local_rank()
+    tp_group = tp_mesh.get_group()
+    pp_group = pp_mesh.get_group()
+
+    logger.info(f"review: {pp_group=}, {tp_group= }")
+
+    logger.info(f"Created device mesh: {device_mesh}\n {tp_mesh=}, {pp_mesh=}\n")
+    # TODO - this assumes 1D mesh, need to update for 2D+ mesh
```
kwen2501, line 267: why is it a 1D mesh?

dist_run.py
```diff
@@ -182,11 +284,12 @@ def main():
 
     # Distribute model on TP mesh
     model.distribute(tp_mesh)
-    logger.info(f"Model: {model}")
+    # logger.info(f"Model: {model}")
```
kwen2501, line 287: temporary code change? (not intended to commit?)

dist_run.py
```diff
@@ -182,11 +284,12 @@ def main():
...
 
-    mbs = 2  # number of micro-batches
+    mbs = 1  # number of micro-batches
```
kwen2501, line 289: temporary code change? (not intended to commit?)

dist_run.py
```diff
@@ -199,7 +302,7 @@ def main():
 
     # Load weights
     logger.info(f"Loading weights for {pp_rank=} on {device=}")
-    with TrackTime("cuda") as timer:
+    with TrackTime() as timer:
```
kwen2501, line 305: I believe this change means that the load itself is synchronous and hence not need for the cuda timer?

dist_run.py
```diff
@@ -212,9 +315,8 @@ def main():
     logger.info(
         f"Stage {rank} has {color.blue}{stage_num_params} params{color.reset}, Size: {color.blue}{stage_size_formatted}{color.reset}\n"
     )
-    
+
     # Setup input position
-    # input_pos for prefill: a list of increasing integers from 0 to seqlen
```
kwen2501, line 217: Is there a reason for removing this comment?

dist_run.py
```diff
@@ -235,31 +337,179 @@ def main():
...
+    # encode the prompt
+    input_ids = _encode_strings(
+        prompt, tokenizer, bos=True, device=device, dtype=torch.int64
+    )
+    logger.info(f"{input_ids[0:8]=}")
+
+    # create a padded tensor for the input prompt
+    padded_sequence, prompt_lengths = _create_padded_prompts(
+        input_ids, tokenizer, seqlen, start_pos, device
+    )
```
kwen2501, line 374: nit: (1) why padding the input_ids requires the tokenizer again? (2) maybe device can be inferred from input_ids?

dist_run.py
```diff
@@ -235,31 +337,179 @@ def main():
...
+    logger.info(f"{prompt_lengths=}")
+    logger.info(f"first prompt {padded_sequence[0, :prompt_lengths[0]+1]=}")
+    if len(prompt_lengths) > 1:
+        logger.info(f"second prompt {padded_sequence[1, :prompt_lengths[1]+1]=}")
```
kwen2501, line 378: nit: can we clean up these logs a little bit? And why do we need the +1?

dist_run.py
```diff
@@ -235,31 +337,179 @@ def main():
...
     logger.info(f"Created schedule: {schedule}")
 
-    with torch.no_grad():  # .inference_mode():
-        if pp_rank == 0:
-            schedule.step(input_ids)
-        else:
-            output = schedule.step()
+    # with CUDATrackTime() as timer:
+    first_pp_group = 0
+    last_pp_group = pp_group_size - 1
```
kwen2501, line 385: Why is "group" a number? Do you mean rank in PP group?

dist_run.py
```diff
@@ -235,31 +337,179 @@ def main():
...
+
+    x_recv = torch.zeros(1, device=device, dtype=torch.int64)
```
kwen2501, line 387: nit: can we add a comment on the "1" part? thx.

dist_run.py
```diff
@@ -235,31 +337,179 @@ def main():
...
+    logger.info(f"{x_recv.shape=}")
```
kwen2501, line 388: nit: remove if not super important?

dist_run.py
```diff
@@ -235,31 +337,179 @@ def main():
...
+                if step < num_tokens - 1:
+                    dist.send(next_token, dst, pp_group)
+                logger.info(
+                    f"SENDING back...from rank={pp_rank} to dst={dst}, data {next_token.shape=}, {next_token=}"
+                )
+                assert next_token != 128006, f"next_token is header id={next_token}"
+
+            else:  # middle pp ranks
+                schedule.step()
+    """
```
kwen2501, line 443: do you mind moving commented-out block to a different PR if not required by this PR?

dist_run.py
```diff
@@ -235,31 +337,179 @@ def main():
...
+    with torch.no_grad():
+        for step in range(num_tokens):
+            # first
+            if pp_rank == 0:
+                schedule.step(padded_sequence)
+                # only receive if not last step
+                if step < num_tokens - 1:
+
```
kwen2501, line 451: nit: remove empty line after `if`

dist_run.py
```diff
@@ -235,31 +337,179 @@ def main():
...
+                output = schedule.step()
+                # need to decode the output
+                decode_results = _batch_decode_next_tokens(
+                    output=output, prompt_lengths=prompt_lengths, tokenizer=tokenizer
+                )
+                logger.info(
+                    f"\n\n{color.green} {'Prefill' if step == 0 else '* Decode *'} responses ====>>>> {color.blue} {decode_results=} \n{color.reset}"
+                )
+
+                next_token = torch.tensor([decode_results[0][0]], device=device)
```
kwen2501, line 472: nit: I am not sure if the use of `torch.tensor` here would cause a sync from GPU to CPU (to get the scalar) then move to the GPU again (to create the tensor). 
If there is no use of `next_token` in CPU domain, better to just use index op here.

Or, is `decode_results` already on CPU? Hmm, then we'd need to think about how to arrange these CPU ops and GPU ops. (Ideally, you would like to fire the send right after step).

dist_run.py
```diff
@@ -235,31 +337,179 @@ def main():
...
+                res.append(decode_results[0][1])
```
kwen2501, line 473: why `next_token` is from [0][0] while `res` is from [0][1]?

dist_run.py
```diff
@@ -235,31 +337,179 @@ def main():
...
+
+                # increment prompt lengths for next token
+                for i in range(len(prompt_lengths)):
+                    prompt_lengths[i] += 1
+                    logger.info(
+                        f"output review {prompt_lengths[i]=}, {padded_sequence[i, prompt_lengths[i]-1]=}"
+                    )
+
+                # logger.info(f"SENDING back...from {rank=} to {dst=}")
+                # logger.info(f"SENDING data {next_token.shape=}, {next_token=}")
```
kwen2501, line 483: nit: remove if not frequently used, or change to logger.debug.

dist_run.py
```diff
@@ -235,31 +337,179 @@ def main():
...
+                        pp_group,
+                    )
+
+            # middle pp ranks
+            else:
+                schedule.step()
+
+        # logger.info(f"REVIEW {padded_sequence[0,:15]=}")
+
+    # logger.info(f"{color.green}Total prefill time: {timer.get_time()} {timer.unit}{color.reset}")
```
kwen2501, line 499: Same

kwen2501 (approved):
Thank for the big effort. LGTM. Would appreciate it if the code can be cleaned up more (please see comments in line)

kwen2501:
Minor comment:
For PR description, would be easier to preserve info in git history if using text than images.

dist_run.py
lessw2020: updated

dist_run.py
lessw2020: updated.

dist_run.py
lessw2020: sure - seqlen is the max len that model can handle or at least the max of what we are running as an available context window (i.e. 4096).
start_pos is where within the sequence we are starting - for a fresh prompt it's 0...but if we start batching things like in splitfuse, then we need to assume we could be continuing from a partial sequence already run earlier...hence why I had the check for a dynamic start point.

dist_run.py
lessw2020: bos = Beginning of Sequence. 
It's the prompt used to start off an interaction, visually it's "<|startoftext|>". 
We put that as the first token for a new prompt submission.

run_dist.sh
lessw2020: hmm, i use it like 50-150x a day during development so I don't want to remove it :) 
the export part is to keep the main gpus free for others that share my server.

dist_run.py
lessw2020: it's not necessarily, it's more that I could not ensure that the return from .size() is capturing all total if 2D+ and there's no len(device_mesh) to quickly get the total participants.
Thus just marked it as a TODO and have been checking with Iris for a better way.

dist_run.py
lessw2020: yes, was just adding to console noise during debugging...reverted.

dist_run.py
lessw2020: this is intentional for now as just focused on single prompt case.  
will update after we get batch support for decoding.

dist_run.py
lessw2020: that's my understanding but to be safe just reverted it back to CUDAtracktime.  We should be careful though as chat uses generic perf_timer atm for their load so if it's not synchronous, we'll not be comparing apples to apples.

dist_run.py
lessw2020: ah sorry, reverted.  I thought metamate had added this (since I have to fight it from randomly adding things when I don't intend for it to) and didn't think two lines of comments was needed here. 
Let's collapse it though as we don't need two lines of "setup input position and then # input_pos for prefill - can just be:
"Setup input_pos for prefill: a list of increasing..."

dist_run.py
lessw2020: I'm not sure - let me open an issue to ensure we resolve this so we can land this for now and then debug it more in depth and ensure best perf.

lessw2020: opened issue: https://github.com/pytorch/torchchat/issues/1147

dist_run.py
lessw2020: this is b/c decode results returns both token_id (int) and token_str (readable).  
This was so we could push the text to the console during processing (and for me, debugging).

dist_run.py
lessw2020: removed

Jack-Khuu:
Rebasing to pick up CI fix

PR #1135, opened 2024-09-12
Improve clone et

larryliu0820:
Check if ExecuTorch is already cloned and no local changes. If
so, skip cloning.

Test Plan: Manually run `install_et.sh` several times and made sure it's
not cloning.


Jack-Khuu (approved):
Simple and sweet, Thanks!

PR #1136, opened 2024-09-12
bump ao to 0.5.0 to enable torchtune across platforms

Gasoonjia:
As torchao 0.5.0 released with mac prebuilt, we update installation script to enable torchtune across different platform

### DIFF
--- a/install/install_requirements.sh
+++ b/install/install_requirements.sh
@@ -95,6 +95,11 @@ LINUX_REQUIREMENTS_TO_INSTALL=(
 )
 
 PLATFORM=$(uname -s)
+
+# Install torchtune and torchao requirements for Linux systems using nightly.
+# For non-Linux systems (e.g., macOS), install torchao from GitHub since nightly
+# build doesn't have macOS build.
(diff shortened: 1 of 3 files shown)


install/install_requirements.sh
```diff
@@ -94,27 +87,10 @@ LINUX_REQUIREMENTS_TO_INSTALL=(
...
-  # For torchao need to install from github since nightly build doesn't have macos build.
-  # TODO: Remove this and install nightly build, once it supports macos
-  (
-    set -x
-    $PIP_EXECUTABLE install git+https://github.com/pytorch/ao.git@e11201a62669f582d81cdb33e031a07fb8dfc4f3
-  )
-fi
+(
+  set -x
+  $PIP_EXECUTABLE install torchao=="0.5.0"
```
Jack-Khuu, line 92: I think it's fine to have it here for now since we might move to nightlies, but this makes more sense in the install_requirements.txt

Jack-Khuu (approved):
Looks great to me

PR #1140, opened 2024-09-13
Torchchat CLI pipeline for Multimodal Models

vmpuri:
Tested using:

```
python3 torchchat.py generate --checkpoint-path ~/path/to/consolidated.pth --tokenizer-path ~/path/to/tokenizer.model --params-path ~/path/to/flamingo.json --image-prompts ~/path/to/dog.jpg --prompt "Is the image of a cat?"
```

Chat does not function properly with image prompts.

### DIFF
--- a/torchchat/cli/builder.py
+++ b/torchchat/cli/builder.py
@@ -15,15 +15,26 @@ import torch
 import torch._dynamo.config
 import torch._inductor.config
 import torch.nn as nn
+
+try:
+    from _torchchat_test_script import flamingo_meta_to_tune
+except ImportError:
+    pass
+
 from distributed import (
     init_distributed,
     launch_distributed,
     ParallelDims,
     parallelize_llama,
 )
+
 from torch.distributed.device_mesh import DeviceMesh
 
-from torchchat.model import Model
+from torchtune.models.convert_weights import meta_to_tune
+
+from torchtune.training import set_default_dtype
+
+from torchchat.model import Model, ModelType
 
 from torchchat.model_config.model_config import resolve_model_config
 from torchchat.utils.build_utils import (
@@ -34,10 +45,6 @@ from torchchat.utils.build_utils import (
 )
 from torchchat.utils.measure_time import measure_time
 from torchchat.utils.quantize import quantize_model
-
-from torchtune.models.convert_weights import meta_to_tune
-
-
 
 
 @dataclass
@@ -143,7 +150,6 @@ class BuilderArgs:
                     if "chat" in path_basename or "instruct" in path_basename:
                         is_chat_model = True
 
-
         output_pte_path = getattr(args, "output_pte_path", None)
         output_dso_path = getattr(args, "output_dso_path", None)
         if output_pte_path and args.dtype.startswith("fast"):
@@ -234,7 +240,12 @@ class TokenizerArgs:
 
         is_tiktoken = self.is_tiktoken
         is_sentencepiece = self.is_sentencepiece
-        use_tiktoken = model.config.transformer_args["text"].use_tiktoken
+        text_args = model.config.transformer_args.get("text")
+        if text_args is None:
+            # TODO: Will be refactored: Currently, the only model that doesn't have text in transfomer_args is Flamingo
+            use_tiktoken = model.config.model_type == ModelType.Flamingo
+        else:
+            use_tiktoken = text_args.use_tiktoken
 
         if not (is_tiktoken == use_tiktoken) or not (is_sentencepiece != use_tiktoken):
             raise RuntimeError(
@@ -266,7 +277,9 @@ class TokenizerArgs:
             raise RuntimeError("cannot find tokenizer model")
 
         if not tokenizer_path.is_file():
-            raise RuntimeError(f"did not find tokenizer at {tokenizer_path}")
+            raise RuntimeError(
+                f"did not find tokenizer at {os.path.abspath(tokenizer_path)}"
+            )
 
         return cls(
             tokenizer_path=tokenizer_path,
@@ -335,7 +348,9 @@ def _load_model_default(builder_args, only_config=False):
 
     if builder_args.params_table and builder_args.params_table.endswith("Tune"):
         print("Loading Tune checkpoint")
-        meta_checkpoint = torch.load(str(builder_args.checkpoint_path), mmap=True, weights_only=True)
+        meta_checkpoint = torch.load(
+            str(builder_args.checkpoint_path), mmap=True, weights_only=True
+        )
         checkpoint = meta_to_tune(meta_checkpoint)
     elif builder_args.checkpoint_dir is not None:
         # Load multiple checkpoint; ignore the single path.
@@ -372,8 +387,17 @@ def _load_model_default(builder_args, only_config=False):
     if "model" in checkpoint and "stories" in str(builder_args.checkpoint_path):
         checkpoint = checkpoint["model"]
 
-    checkpoint = {"model." + k: v for k, v in checkpoint.items()}
-    model.load_state_dict(checkpoint, assign=True, strict=True)
+    if model.config.model_type == ModelType.Flamingo:
+        # TODO: Refactor this. For now, overwrite the model with model loaded from params_path
+        with set_default_dtype(builder_args.precision), torch.device(
+            builder_args.device
+        ):
+            model = Model.from_params(builder_args.params_path)
+        state_dict = flamingo_meta_to_tune(checkpoint)
+        model.model.load_state_dict(state_dict)
+    else:
+        checkpoint = {"model." + k: v for k, v in checkpoint.items()}
+        model.load_state_dict(checkpoint, assign=True, strict=True)
 
     return model
 
--- a/torchchat/generate.py
+++ b/torchchat/generate.py
@@ -12,12 +12,26 @@ import time
 
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
+from os import PathLike
 from pathlib import Path
-from typing import List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
 
 import torch
 import torch._dynamo.config
 import torch._inductor.config
+
+try:
+    from _torchchat_test_script import flamingo_transform, padded_collate
+except ImportError:
+    pass
+
+from PIL import Image
+
+# torchtune model definition dependencies
+from torchtune.data import Message
+from torchtune.generation._generation import sample as tune_sample
+from torchtune.models.llama3 import llama3_tokenizer
+from torchtune.training import set_default_dtype
 
 from torchchat.cli.builder import (
     _initialize_model,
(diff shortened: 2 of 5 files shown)


torchchat/generate.py
```diff
@@ -148,9 +168,16 @@ def from_args(cls, args):
         pte_path = getattr(args, "pte_path", None)
         sequential_prefill = args.sequential_prefill or bool(dso_path) or bool(pte_path)
 
+        # Validate that all image prompts exist before expensive model load
+        image_prompts = getattr(args, "image_prompts", None)
+        if image_prompts:
```
Jack-Khuu, line 173:
```suggestion
        if image_prompts := getattr(args, "image_prompts", None):
```

torchchat/generate.py
```diff
@@ -300,9 +328,9 @@ def sample(
         self,
         logits,
         need_probs: bool,
-        temperature: float = 1.0,
+        temperature: float = 0,
```
Jack-Khuu, line 331: For prosperity, let's avoid changing this in this diff

torchchat/generate.py
```diff
@@ -328,6 +357,10 @@ def prefill(
                 x_sliced, ip_sliced = x[:, i].view(-1, 1), input_pos[i].view(-1)
                 # logging.debug(f"<sliced> x: {x_sliced}, input_pos: {ip_sliced}")
                 logits = model(x_sliced, ip_sliced)  # (x[:, i], input_pos[i])
+        elif batch is not None:
+            logits = model(**batch)[:, -1]
+            print("logits", logits)
```
Jack-Khuu, line 362: Accidental print?

torchchat/generate.py
```diff
@@ -12,21 +12,30 @@
...
 import torch
 import torch._dynamo.config
 import torch._inductor.config
 
+from _torchchat_test_script import flamingo_transform, padded_collate
+from PIL import Image
+
+# torchtune model definition dependencies
+from torchtune.data import Message
+from torchtune.generation._generation import sample
```
Jack-Khuu, line 28:
```suggestion
from torchtune.generation._generation import sample as tune_sample
```
To disambiguate

torchchat/generate.py
```diff
@@ -328,6 +357,10 @@ def prefill(
                 x_sliced, ip_sliced = x[:, i].view(-1, 1), input_pos[i].view(-1)
                 # logging.debug(f"<sliced> x: {x_sliced}, input_pos: {ip_sliced}")
                 logits = model(x_sliced, ip_sliced)  # (x[:, i], input_pos[i])
+        elif batch is not None:
+            logits = model(**batch)[:, -1]
+            print("logits", logits)
+            return sample(logits, 0, 500)
```
Jack-Khuu, line 363:
```suggestion
            return tune_sample(logits, 0, 500)
```

torchchat/generate.py
```diff
@@ -328,6 +357,10 @@ def prefill(
                 x_sliced, ip_sliced = x[:, i].view(-1, 1), input_pos[i].view(-1)
                 # logging.debug(f"<sliced> x: {x_sliced}, input_pos: {ip_sliced}")
                 logits = model(x_sliced, ip_sliced)  # (x[:, i], input_pos[i])
+        elif batch is not None:
```
Jack-Khuu, line 360: I don't think this MM plays well with sequential_prefill. Can we bump this conditional up before that check and call out that seq_prefill is not tested with MM

torchchat/generate.py
```diff
@@ -517,26 +572,37 @@ def generate(
...
+                ):
                     model.setup_caches(max_batch_size=1, dtype=self.dtype)
                 else:
                     model.setup_caches(max_batch_size=1, max_seq_length=max_seq_length)
                 if is_speculative and draft_model is not model:
                     draft_model.setup_caches(
-                        max_batch_size=1, max_seq_length=max_seq_length
+                        max_batch_size=1,
+                        max_seq_length=max_seq_length,
+                        dtype=torch.bfloat16,
```
Jack-Khuu, line 595:
```suggestion
```

Debug remnant

torchchat/generate.py
```diff
@@ -517,26 +572,37 @@ def generate(
...
                     )
+            if model.config.model_type == ModelType.Flamingo:
+                model.reset_caches()
```
Jack-Khuu, line 598: Why?

torchchat/generate.py
```diff
@@ -517,26 +572,37 @@ def generate(
...
 
         # create an empty tensor of the expected final shape and
         # fill in the current tokens
         empty = torch.empty(T_new, dtype=dtype, device=device)
+
```
Jack-Khuu, line 603: Remove added white space

torchchat/generate.py
```diff
@@ -610,14 +680,9 @@ def generate(
...
             ):
-                generated_tokens.append(generated_token)
+                generated_tokens.append(generated_token.view(-1))
                 yield generated_token, None
 
-            seq[T + 1 : T + 1 + len(generated_tokens)] = torch.cat(generated_tokens)
-            seq = seq[
-                : T + 1 + len(generated_tokens)
-            ]  # If we dont generate all the way to max_new_tokens slice off the extra space we allocated.
-
```
Jack-Khuu, line 620: why?

torchchat/generate.py
```diff
@@ -655,10 +720,36 @@ def chat(
     ):
         if generator_args.chat_mode:
             print("Starting Interactive Chat")
-        encoded = self.encode_tokens(
-            generator_args.prompt, bos=True, device=self.builder_args.device
-        )
-        logging.debug(encoded)
+
+        if generator_args.image_prompts is not None:
+            print("image prompts", generator_args.image_prompts)
```
Jack-Khuu, line 725:
```suggestion
            print("Image prompts: ", generator_args.image_prompts)
```

torchchat/model.py
```diff
@@ -195,11 +197,15 @@ def from_params(cls, params_path):
             # try to interpret as a single transformer config
             transformer_args: Dict[str, TransformerArgs] = {}
             transformer_args["text"] = TransformerArgs.from_params(loaded_params)
-            model_type = ModelType.TextOnly
+            if model_type := loaded_params.get("model_type", None) is None:
```
Jack-Khuu, line 200:
```suggestion
            if (model_type := loaded_params.get("model_type", None)) is None:
```

torchchat/generate.py
```diff
@@ -105,6 +122,9 @@ class GeneratorArgs:
         None  # When passed into the Generator, this will be used as the system prompt
     )
     encoded_prompt: Optional[torch.Tensor] = None
+    image_prompts: Optional[List[Union[str, PathLike, bytes]]] = (
```
Gasoonjia, line 125: Let's use sequence instead of List for better clarity:

```suggestion
    image_prompts: Optional[Sequence[Union[str, PathLike, bytes]]] = (
```

torchchat/generate.py
Gasoonjia: setup_caches rn of draft model should always be setup_caches(max_batch_size, max_seq_length), will not have dtype

torchchat/model.py
```diff
@@ -741,10 +770,9 @@ def apply_rotary_emb(x: Tensor, freqs_cis: Tensor) -> Tensor:
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 try:
-    from executorch.extension.pybindings import portable_lib as exec_lib
-    
     # ET changed the way it's loading the custom ops so it's not included in portable_lib but has to be loaded separately.
     from executorch.examples.models.llama2.custom_ops import sdpa_with_kv_cache  # no-qa
+    from executorch.extension.pybindings import portable_lib as exec_lib
```
Gasoonjia, line 775: This will introduce issue previously -- we have to import exec_lib first then sdpa_with_kv_cache

Gasoonjia (commented):
I think we can add flamingo as a new model name in the codebase to make the command easier. That's is, instead of current cmd:

```
python3 torchchat.py generate --checkpoint-path ~/path/to/consolidated.pth --tokenizer-path ~/path/to/tokenizer.model --params-path ~/path/to/flamingo.json --image-prompts ~/path/to/dog.jpg --prompt "Is the image of a cat?"
```

maybe we can have something like:
```
python3 torchchat.py generate flamingo --checkpoint-path ~/path/to/consolidated.pth --image-prompts ~/path/to/dog.jpg --prompt "Is the image of a cat?"
```

torchchat/generate.py
vmpuri: code is unused.

torchchat/cli/builder.py
```diff
@@ -234,7 +239,12 @@ def validate_model(
 
         is_tiktoken = self.is_tiktoken
         is_sentencepiece = self.is_sentencepiece
-        use_tiktoken = model.config.transformer_args["text"].use_tiktoken
+        text_args = model.config.transformer_args.get("text")
+        if text_args is None:
+            # TODO: Will be refactored: Currently, the only model that doesn't have text in transfomer_args is Flamingo
+            use_tiktoken = model.config.model_type == ModelType.Flamingo
```
Jack-Khuu, line 245: Just calling out that this might cause issues with other use cases that manually pass in checkpoints/params

Not something that needs to be addressed in this diff, but may need to revisit sooner rather than later

torchchat/generate.py
Jack-Khuu: nit: not required

torchchat/generate.py
Jack-Khuu: Legit

torchchat/generate.py
Jack-Khuu: Legit

torchchat/generate.py
Jack-Khuu: nit: Not required

torchchat/generate.py
Jack-Khuu: nit: not required

torchchat/generate.py
Jack-Khuu: @Gasoonjia Do you know why we set then reset?

Jack-Khuu (approved):

torchchat/model.py
Jack-Khuu: We need bump back up, or the imports fail 

We should comment this so people know in the future

torchchat/cli/builder.py
```diff
@@ -15,15 +15,26 @@
 import torch._dynamo.config
 import torch._inductor.config
 import torch.nn as nn
+
```
Gasoonjia, line 18: plz add a comment to describe why we need to pass the import error

torchchat/cli/builder.py
Gasoonjia: The reason why we need this is on my side I didn't provide a unify config system: the way we get model config from model between tune-support and chat-support is different.
I've put this into my BE doc, and will solve it whenever we can breath.

PR #1142, opened 2024-09-13
modified generate.py for cli and browser

nobelchowdary:
This allows to run the server and generate chat versions seamlessly


vmpuri:
Is the role ever *not* a string? If that's the case, we should find the source of this bug rather than adding a typecast here.

Jack-Khuu:
Closing outdated PR

PR #1148, opened 2024-09-15
[Distributed] Accept model names

kwen2501:
### Change 1
Add argparser to accept model name and other arguments (e.g. PP degree).

```
$ torchrun dist_run.py -h
usage: dist_run.py [-h] [--pp PP] {llama2-7b-chat,llama3}

positional arguments:
  {llama2-7b-chat,llama3}
                        Name of the model to load

options:
  -h, --help            show this help message and exit
  --pp PP               Pipeline parallel degree
```

Supported model names:
```
llama2-7b-chat
llama3
```
(Following model aliases in [models.json](https://github.com/pytorch/torchchat/blob/main/torchchat/model_config/models.json))

### Change 2
Renamed variable `hf_model_name` to `distribution`.

A distribution looks like this:
`meta-llama/Llama-2-7b-chat-hf`
`meta-llama/Meta-Llama-3-8B-Instruct`

### Change 3
Fix `output` undefined error when PP = 1.

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -4,10 +4,11 @@
 # This source code is licensed under the license found in the
 # LICENSE file in the root directory of this source tree.
 
+import argparse
 import os
 from pathlib import Path
 from types import SimpleNamespace
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 # Run command:
 # torchrun --nproc-per-node 4 dist_run.py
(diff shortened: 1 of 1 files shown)


dist_run.py
```diff
@@ -182,7 +195,8 @@ def main():
 
     # Distribute model on TP mesh
     model.distribute(tp_mesh)
-    logger.info(f"Model: {model}")
+    if rank == 0:
```
lessw2020, line 198: nit: already filtering based on ranks in the dist_run.sh launcher rather than having express rank based logging in the code.

lessw2020 (approved):
looks good! 
I also have toml config supporting code in distributed/config_manager.py that I had integrated earlier, so we should add in both cmd line and toml support ultimately (similar to titan where you can use either).

Jack-Khuu:
Rebasing on main to pick up CI fix

dist_run.py
kwen2501: Yeah, it is a hard choice.. I don't know if regular users use filter (I never remember that option's name) and want to have a minimal command line work.

kwen2501:
> Rebasing on main to pick up CI fix

Thanks @Jack-Khuu !

PR #1149, opened 2024-09-15
[Distributed] fix pp=1 case; clean up

kwen2501:
When there is only 1 PP stage, we can skip the sendrecv.

Plus some cleanup and renaming.

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -135,8 +135,8 @@ def _load_model_weights(stage_module, distribution, device, model_config):
 def _encode_strings(
     strings: List[str],
     tokenizer,
-    bos: bool = True,
-    device: torch.device = "cuda:0",
+    bos: bool,
+    device: torch.device,
     dtype=torch.int64,
 ) -> List[torch.Tensor]:
     """Encode a list of prompt strings into a list of tensor token ids."""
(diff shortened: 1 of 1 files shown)


dist_run.py
```diff
@@ -388,88 +377,74 @@ def main(args):
...
-                        next_token,
-                        dst,
-                        pp_group,
-                    )
+            # sendrecv between last and first ranks, only if:
+            # first_pp_rank != last_pp_rank.
+            if pp_rank == last_pp_rank and pp_rank != first_pp_rank:
+                dist.send(
+                    new_token,
+                    dst=dist.get_global_rank(pp_group, first_pp_rank),
```
lessw2020, line 423: dst and src shouldn't be getting recreated every iter, since they don't change on a per iter basis. 
This is why I had moved them out of the loop previously. 
Not sure how expensive the dist.get_global_rank is but no need imo to be calling it over and over here.

dist_run.py
```diff
@@ -388,88 +377,74 @@ def main(args):
...
+                    group=pp_group,
+                )
+            elif pp_rank == first_pp_rank and pp_rank != last_pp_rank:
+                dist.recv(
+                    new_token,
+                    src=dist.get_global_rank(pp_group, last_pp_rank),
```
lessw2020, line 429: same as above w/dst, why do we call this api over and over to get src within the loop instead of once out of the loop?

lessw2020 (approved):
looks good!
Flagged one item re: why dst and scr got moved from out of loop, to in loop... seems no reason to recreate dst and src in every iter vs set once and keep them out of the loop. But, it's minor and prefer to land this now as planning to work on batch decoding today.

dist_run.py
kwen2501: Makes sense. Will upload a PR to improve it. Thanks!

PR #1150, opened 2024-09-15
[Distributed] Move global id calculation

kwen2501:
Accommodate review comments of previous PR:
avoid calculating these IDs multiple times.


lessw2020 (approved):
looks great! 
Note that I had already done this in my batch decoding PR, but I updated my variable names to match yours for dst/src. 
Thus, may not need to merge this but approving regardless.

PR #1151, opened 2024-09-15
[distributed] add batch decoding 

lessw2020:
This PR:
1.  enables batch decoding such that multiple prompts can be run and processed at the same time. 
2. moves the dst/src calcs for pp comms out of the decoding loop
3. updates final display to show each respective prompt and result. 
4. Tested with both PP=1 and PP=2

Screenshot 2024-09-15 at 12 48 00 PM

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -219,10 +219,11 @@ def _update_padded_sequence(
     new_token: torch.Tensor,
     prompt_lengths: List[int],
 ) -> None:
-    # TODO: this is a hacky way to update the padded sequence: when there is
-    # more than one prompt, the for loop and the assignment is incompatible.
+
     for i in range(len(prompt_lengths)):
-        padded_sequence[i, prompt_lengths[i]] = new_token
+        prompt_lengths[i] += 1
+        padded_sequence[i, prompt_lengths[i] - 1] = new_token[i, 0]
+        logger.info(f"updated prompt {i} with new token {new_token[i, 0]}")
 
 
 def _cleanup():
@@ -242,7 +243,7 @@ def main(args):
     distribution, model_dtype = NAME_TO_DISTRIBUTION_AND_DTYPE[model_name]
     logger.info(f"Using HF model weights from {distribution} and dtype {model_dtype}")
 
-    config = ModelArgs.from_name(distribution).transformer_args['text']
+    config = ModelArgs.from_name(distribution).transformer_args["text"]
     logger.info(f"Chat Model Config: {config}")
 
     tokenizer = _build_chat_tokenizer(model_name)
@@ -295,7 +296,7 @@ def main(args):
         logger.info(f"Model: {model}")
 
     mbs = 1  # number of micro-batches
-    mb_size = 1  # micro-batch size
+    mb_size = 2  # micro-batch size
     batch_size = mbs * mb_size  # total batch size
 
     seqlen = 4096  # sequence length
@@ -343,6 +344,7 @@ def main(args):
 
     prompt = [
         "What is snow?",
+        "Where does Santa Claus live?",
     ]
 
     """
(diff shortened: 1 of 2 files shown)


dist_run.py
```diff
@@ -219,10 +218,10 @@ def _update_padded_sequence(
     new_token: torch.Tensor,
     prompt_lengths: List[int],
 ) -> None:
-    # TODO: this is a hacky way to update the padded sequence: when there is
-    # more than one prompt, the for loop and the assignment is incompatible.
     for i in range(len(prompt_lengths)):
-        padded_sequence[i, prompt_lengths[i]] = new_token
+        prompt_lengths[i] += 1
```
kwen2501, line 222: No need to increment `prompt_length` here. It was done for every rank after the step now. See below.

dist_run.py
```diff
@@ -412,40 +421,45 @@ def main(args):
...
             # Update input sequence with new token
             if pp_rank == first_pp_rank:
-                _update_padded_sequence(
-                    padded_sequence, new_token, prompt_lengths
-                )
+                _update_padded_sequence(padded_sequence, new_token, prompt_lengths)
 
-            # increment prompt lengths for next token
-            for i in range(len(prompt_lengths)):
-                prompt_lengths[i] += 1
```
kwen2501, line 441: See comment above. `prompt_lengh` is incremented here, every rank would do so. And only once.

dist_run.py
```diff
@@ -412,40 +421,45 @@ def main(args):
...
-                res.append(decode_results[0][1])
+                for i in range(len(decode_results)):
+                    res[i].append(decode_results[i][1])
+                    new_token[i, 0] = torch.tensor(
+                        [decode_results[i][0]], device=device
+                    )  # decode_results[i][0]
+
+                # increment prompt lengths for next token
+                for i in range(len(prompt_lengths)):
+                    prompt_lengths[i] += 1
```
kwen2501, line 432: I moved this out (see below) in the "fix pp=1" PR. The move is part of the fix.

dist_run.py
```diff
@@ -412,40 +421,45 @@ def main(args):
...
+    # Display the decoding results
 
     # output formatted response via last pp group and tp rank 0
     if pp_rank == last_pp_rank and tp_rank == 0:
-        logger.info(f"Prompt:{color.green} {prompt[0]} {color.reset}")
-        formatted_response = " ".join(res)
-        logger.info(f"$$$$$$ {color.blue}{formatted_response} {color.reset}  $$$$$")
+        for i in range(len(prompt_lengths)):
+            logger.info(f"\nPrompt:{color.green} {prompt[i]} {color.reset}")
+            formatted_response = "".join(res[i])
```
kwen2501, line 459: `"".join` does not work for llama2 (don't know why yet), that's why I changed it to `" "`.

kwen2501:
If `pp_degree` = 1, the change in this PR would cause `prompt_length` to be incremented twice per iteration.

That's why in PR #1149 I consolidated the increments in one place.

dist_run.py
lessw2020: I see was wondering why you did that. 
I'll add an if/else then, as we can't wreck the formatting of llama3 outputs to make llama2 work.

lessw2020: added if/else based on model name

dist_run.py
lessw2020: removed as part of an improved solution for both pp=1 and pp=2.

dist_run.py
lessw2020: got it - I thought this was breaking pp=2 but it was a different issue. reverted.

dist_run.py
lessw2020: got it - I thought this was breaking pp=2 but it was a different issue. reverted.

kwen2501 (approved):
LGTM!

PR #1152, opened 2024-09-16
UI and API implementation for base64 encoded image input

vmpuri:
As described, tested with a basic image input.

### DIFF
--- a/torchchat/generate.py
+++ b/torchchat/generate.py
@@ -364,6 +364,8 @@ class Generator:
                 x_sliced, ip_sliced = x[:, i].view(-1, 1), input_pos[i].view(-1)
                 # logging.debug(f"<sliced> x: {x_sliced}, input_pos: {ip_sliced}")
                 logits = model(x_sliced, ip_sliced)  # (x[:, i], input_pos[i])
+        elif self.model.config.model_type == ModelType.Flamingo:
+            logits = model(x)
         else:
             # input_pos: [B, S]
             logits = model(x, input_pos)
@@ -383,11 +385,14 @@ class Generator:
     ) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
         # input_pos: [B, 1]
         assert input_pos.shape[-1] == 1
-        if model.config.model_type == ModelType.Flamingo and batch is not None:
-            x = x.view(1, -1)
-            logits = model(x, encoder_mask=batch["encoder_mask"][:, -1:])
+        x = x.view(1, -1)
+        if model.config.model_type == ModelType.Flamingo:
+            if batch is not None:
+                logits = model(x, encoder_mask=batch["encoder_mask"][:, -1:])
+            else:
+                logits = model(x)
         else:
-            logits = model(x.view(1, -1), input_pos)
+            logits = model(x, input_pos)
         # print(f"x: {x},\n  input_pos: {input_pos}\n")
         return self.sample(logits, need_probs=need_probs, **sampling_kwargs)
 
--- a/torchchat/usages/browser.py
+++ b/torchchat/usages/browser.py
@@ -1,5 +1,13 @@
+import base64
+import logging
 import time
+from pathlib import Path
+
 import streamlit as st
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+
 from openai import OpenAI
 
 st.title("torchchat")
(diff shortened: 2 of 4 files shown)


torchchat/usages/browser.py
```diff
@@ -1,39 +1,123 @@
...
+
 from openai import OpenAI
 
 st.title("torchchat")
 
 start_state = [
     {
         "role": "system",
-        "content": "You're an assistant. Answer questions directly, be brief, and have fun.",
+        "content": "You're a helpful assistant - have fun.",
```
Gasoonjia, line 18: Why do we need to update the system prompt? Any source of truth for that?

torchchat/generate.py
```diff
@@ -383,11 +385,14 @@ def decode_one_token(
...
-            logits = model(x, encoder_mask=batch["encoder_mask"][:, -1:])
+        x = x.view(1, -1)
+        if model.config.model_type == ModelType.Flamingo:
+            if batch is not None:
+                logits = model(x, encoder_mask=batch["encoder_mask"][:, -1:])
+            else:
+                logits = model(x)
         else:
-            logits = model(x.view(1, -1), input_pos)
+            logits = model(x, input_pos)
```
Gasoonjia, line 395: Curious why do we need to update generation logic? I thought only the input encoding method has been changed right?

torchchat/usages/server.py
```diff
@@ -125,4 +125,5 @@ def initialize_generator(args) -> OpenAiApiGenerator:
 
 def main(args):
     app = create_app(args)
+    # app.run(host="::", port=8085)
```
Gasoonjia, line 128: remove comment?

torchchat/usages/browser.py
vmpuri: No source of truth - I just made it shorter to reduce the amount of pre-filling the model needs to do. I have another PR that allows users to change this.

torchchat/usages/openai_api.py
```diff
@@ -4,17 +4,23 @@
...
 
 from abc import ABC
 from dataclasses import dataclass
+from io import BytesIO
 from pwd import getpwuid
 from typing import Any, Dict, List, Optional, Union
 
 import torch
 
+from _torchchat_test_script import flamingo_transform, padded_collate
```
Gasoonjia, line 20: maybe we need to add try and except here?

torchchat/generate.py
vmpuri: There's a bug in generate that causes the model to error out with a non-image input prompt. The fix is a separate commit in this PR, but I can split it into another if needed.

Gasoonjia: keep here is fine i think😊

torchchat/usages/browser.py
```diff
@@ -53,15 +137,20 @@ def get_streamed_completion(completion_generator):
...
-                client.chat.completions.create(
-                    model="llama3",
-                    messages=st.session_state.messages,
-                    max_tokens=response_max_tokens,
-                    stream=True,
+        try:
+            response = st.write_stream(
+                get_streamed_completion(
+                    client.chat.completions.create(
+                        model="llama3",
```
Jack-Khuu, line 144: remind me why this is set to llama3?

torchchat/usages/openai_api.py
```diff
@@ -232,18 +278,55 @@ def __init__(self, *args, **kwargs):
...
+                + 1
+                if self.draft_model is not None
+                else self.model.config.transformer_args["text"].max_seq_length
+            )
         # The System fingerprint is a unique identifier for the model and its configuration.
         self.system_fingerprint = (
             f"{self.builder_args.device}_{self.builder_args.precision}"
         )
 
+    def _openai_messages_to_torchtune(self, messages: List[_AbstractMessage]):
```
Jack-Khuu, line 295: nit: return type

torchchat/generate.py
```diff
@@ -364,6 +364,8 @@ def prefill(
                 x_sliced, ip_sliced = x[:, i].view(-1, 1), input_pos[i].view(-1)
                 # logging.debug(f"<sliced> x: {x_sliced}, input_pos: {ip_sliced}")
                 logits = model(x_sliced, ip_sliced)  # (x[:, i], input_pos[i])
+        elif self.model.config.model_type == ModelType.Flamingo:
```
Jack-Khuu, line 367: Why did the previous PR work without this line?

Jack-Khuu (approved):

torchchat/generate.py
vmpuri: It did not, must have missed doing a full grid of test before pushing.

torchchat/usages/browser.py
vmpuri: For now, this could be set to anything - I just set the string so the API doesn't complain. 

Since loading a model is very expensive, we just load whatever the server is launched with. 

I could add a task for selecting a model from a list since the models endpoint is already implemented.

Jack-Khuu: Not in this PR, but we should comment this

torchchat/usages/server.py
Jack-Khuu: lmao

Jack-Khuu (approved):

PR #1153, opened 2024-09-16
[1/n llava]unify model construction ppl

Gasoonjia:
This PR adopts the same pipeline to construct both chat model and tune model; previously we used TransformerArgs to construct chat-backend model while dictionary for tune-backend model. 
Also fix some annoying hacky stuff for configuration.


distributed/parallelize_llama.py
```diff
@@ -62,7 +62,7 @@ def apply_tp(
     # after we apply TP to the model. Because we don't want to change model code 
     # when applying TP. We need to have change to ensure KVCache has the correct
     # size as k and v.
-    model.config.transformer_args["text"].n_local_heads = model.config.transformer_args["text"].n_local_heads // tp_mesh.size()
+    model.model.config.n_local_heads = model.model.config.n_local_heads // tp_mesh.size()
```
Jack-Khuu, line 65: `model.model` is really hard to reason about... 

The former was clunky, but model.model is worse

Jack-Khuu: "text" is not sustainable either, especially if the number of modules increases. It needs fixing, but this might not be perfectly there yet

torchchat/cli/builder.py
```diff
@@ -240,12 +240,7 @@ def validate_model(
 
         is_tiktoken = self.is_tiktoken
         is_sentencepiece = self.is_sentencepiece
-        text_args = model.config.transformer_args.get("text")
-        if text_args is None:
-            # TODO: Will be refactored: Currently, the only model that doesn't have text in transfomer_args is Flamingo
-            use_tiktoken = model.config.model_type == ModelType.Flamingo
-        else:
-            use_tiktoken = text_args.use_tiktoken
+        use_tiktoken = model.config.use_tiktoken
```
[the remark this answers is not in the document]
Gasoonjia: I think at most one text-modality transformer should be part of assumption.
After recent harsh time, a north-star doc should have about what's the assumption of our model will be, and what's the target hierarchy will be.

distributed/parallelize_llama.py
Gasoonjia: That's annoying, i'm 100% agree.
I will remove `model.model` as soon as I can.

torchchat/model.py
```diff
@@ -399,8 +397,9 @@ def reset_caches(self):
 
 
 class Transformer(nn.Module):
-    def __init__(self, config: TransformerArgs) -> None:
+    def __init__(self, config: Dict[str, Any]) -> None:
```
Jack-Khuu, line 400: Not a fan of this one, Transformer taking TransformerArgs is the most intuitive set up and matches the other classes

Gasoonjia: SG. Bring it back

torchchat/model.py
```diff
@@ -163,50 +163,62 @@ def from_params(cls, params):
...
+
     model_type: ModelType
-    transformer_args: Dict[str, Union[Dict, TransformerArgs]]
+    transformer_args: Dict[str, Dict[str, Any]]
+    use_tiktoken: bool
 
     def __init__(
         self,
-        transformer_args: Union[TransformerArgs, Dict[str, TransformerArgs]],
+        transformer_args: Dict[str, Dict[str, Any]],
```
Jack-Khuu, line 184: We should find a way to reconcile `Dict[str, Any]` into a TransformerArgs in a future PR

This makes this work well since we have 3 "cases", but storing/passing around an untyped Dict makes me nervous

torchchat/model.py
```diff
@@ -292,6 +304,7 @@ def __init__(self, config: ModelArgs) -> None:
         super().__init__()
         self.config = config
         self.model = self.build_model()
+        self.text_transformer_args = None
```
Jack-Khuu, line 307: Comment on this since it is a special case

Jack-Khuu (approved):

torchchat/generate.py
```diff
@@ -814,15 +810,9 @@ def chat(
...
-            text_transformer_args = self.model.config.transformer_args.get("text", None)
+        elif not generator_args.is_torchtune_model:
             max_seq_length = min(
-                encoded.size(0) + generator_args.max_new_tokens,
-                (
-                    text_transformer_args.block_size
-                    if text_transformer_args is not None
-                    else 2048
-                ),
+                encoded.size(0) + generator_args.max_new_tokens, max_seq_length
```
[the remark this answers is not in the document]
Gasoonjia: good catch! not sure why this happen, probaly a typo. Will fix it.

torchchat/model.py
```diff
@@ -163,50 +163,62 @@ def from_params(cls, params):
 
 @dataclass
 class ModelArgs:
+    """
+    A data class to describe the structure of a model.
+    Attributes:
+        model_type (ModelType): The type of the model. This attribute is used to categorize the model into different classes.
+        transformer_args (Dict[str, Dict[str, Any]]): A dictionary containing the parameters for each transformer in the model.
+            The outer dictionary has transformer names as keys and inner dictionaries as values. Each inner dictionary contains
+            the parameter names and their corresponding values for the respective transformer.
```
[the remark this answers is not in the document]
Gasoonjia: for unification. this arg takes charge for describing architecture for all models, including tune-backends, chat-backends, and even mix-backends. so we need a unify way to descible how we will set up them.
for chat-backend modules, the inner Dict will be converted into tranformerArg afterwards.

torchchat/model.py
Gasoonjia: More than agree. My mental model would be creating an abstract class containig essential apis for all module configurations, and for different transformer (e.g. ours, tunes, etc) we have a different implementation. Dict[str, Any] is not a great way.
Let me add some comments in our codebase to highlight that.

PR #1154, opened 2024-09-17
[Distributed] Enable KV cache

kwen2501:
Previously KV cache was turned off in distributed cases, enabling it now.

The cache size is related to TP degree -- because the heads are divided across ranks.

Also lowered KV cache instantiation from root level to attention layer (which actually owns the cache).


torchchat/model.py
```diff
@@ -442,8 +442,11 @@ def setup_caches(self, max_batch_size, max_seq_length):
         self.max_seq_length = max_seq_length
         self.max_batch_size = max_batch_size
         for b in self.layers.values():
-            b.attention.kv_cache = KVCache(
-                max_batch_size, max_seq_length, self.config.n_local_heads, head_dim
```
Jack-Khuu, line 446: This is fine, just calling it out as a note to self

Local `head_dim` is defined as `self.config.dim // self.config.n_heads`. 

In Attention, head_dim is pulled from `config.head_dim`. Recall that TransformerArgs (config) is also initialized as `self.config.dim // self.config.n_heads`

torchchat/model.py
```diff
@@ -442,8 +442,11 @@ def setup_caches(self, max_batch_size, max_seq_length):
...
+            # Lower the setup_cache call to the attention module because tensor
```
Jack-Khuu, line 445: head_dim as defined on 440 is no longer used, let's remove it

Jack-Khuu (approved):

torchchat/model.py
kwen2501: Good point, thanks!

lessw2020 (approved):
Thanks for enabling, looks good!

PR #1156, opened 2024-09-17
Fix MPS Test by forcing a PT version

Jack-Khuu:
Flakey tests depending on which version of Python is used. Needs > 3.10.0


Gasoonjia (approved):

Jack-Khuu:
Testing idea, but not working, will need to dig into the tests later

PR #1159, opened 2024-09-17
[aoti] Remove need for -l in cmake call

angelayi:
Removes the need for the `-l` in the cmake call by storing the tokenizer type during export time in the PT2.
Stacked on top of https://github.com/pytorch/torchchat/pull/896

### DIFF
--- a/torchchat/export.py
+++ b/torchchat/export.py
@@ -5,13 +5,13 @@
 # LICENSE file in the root directory of this source tree.
 
 import os
-from typing import Optional
+from typing import Dict, Optional
 
 import torch
+import torch._inductor
 import torch.nn as nn
 
 from torch.export import Dim
-import torch._inductor
 
 from torchchat.cli.builder import (
     _initialize_model,
@@ -39,6 +39,7 @@ def export_for_server(
     output_path: str = "model.pt2",
     dynamic_shapes: bool = False,
     package: bool = True,
+    metadata: Optional[Dict[str, str]] = None,
 ) -> str:
     """
     Export the model using AOT Compile to get a .dso for server use cases.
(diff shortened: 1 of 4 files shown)


Jack-Khuu:
@angelayi The rebase on this looks like a rough, wanna recreate?

torchchat/export.py
```diff
@@ -39,6 +39,7 @@ def export_for_server(
     output_path: str = "model.pt2",
     dynamic_shapes: bool = False,
     package: bool = True,
+    metadata: Optional[Dict[str, str]] = None,
```
Jack-Khuu, line 42:
```suggestion
    metadata: Dict[str, str] = {},
```

We can delete `metadata = metadata or {} `

Jack-Khuu (approved):

torchchat/export.py
```diff
@@ -78,11 +78,12 @@ def export_for_server(
...
             options=options,
         )
-        
+
         if package:
             from torch._inductor.package import package_aoti
+
             path = package_aoti(output_path, path)
-    
+
```
angelayi, line 86: sorry, somehow my editor added on a bunch of formatting changes here.. hope it's not too confusing otherwise I can try to remove them.

Jack-Khuu: No worries, these are good lint fixes

Jack-Khuu (approved):
Legit thanks for the updates

torchchat/export.py
mikekgfb: general advice is to not have mutable structures as default args because they survive invocations =>
https://docs.python-guide.org/writing/gotchas/

Jack-Khuu (approved):

PR #1160, opened 2024-09-18
[Distributed] Use Tensor Parallel instead of Sequence Parallel

kwen2501:
SP is one step further than TP in that it further distributes the layer norm computation.

There are a couple reason to turn the dial back to pure TP:
(1) LLM inference has a prefill phase and a decoding phase which have different seqlen. The decoding phase has a seqlen of 1, to which SP cannot be applied. We don't want to create two models and apply SP and TP separately.
(2) The major motivation of SP is to reduce activation envelope. While this is important in training (bc backward needs those intermediates), we are using `torch.no_grad()` in inference, in which case activations are not kept anyway.
(3) While it is true that AllReduce = AllGather + ReduceScatter, in small-size region latency dominates, so launching 1 collective may be better than launching two collectives. 

- Removed SP feature from model
- Removed `// sp_degree` in PP activation size

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -258,12 +258,11 @@ def main(args):
     assert world_size % pp_degree == 0
     assert config.n_layers % pp_degree == 0
 
-    # Sequence parallel is enabled in this program
-    # Sequence parallel = Tensor parallel + dividing sequence by tp_degree at layer boundary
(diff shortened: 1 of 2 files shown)


lessw2020 (approved):
Thanks for the detailed justification for this. Looks good!

PR #1161, opened 2024-09-18
[Distributed] Follow upstream TransformerArgs changes

kwen2501:
#1153 made a change that `ModelArgs.transformer_args["text"]` is now a dict. 

We need to convert the dict into `TransformerArgs` using `TransformerArgs.from_param(...)`.

Potential TODO:
Create model instead of creating the inner Transformer.


lessw2020 (approved):
Thanks for updating.

PR #1162, opened 2024-09-18
[Distributed] Separate prefill and decode

kwen2501:
Prefill phase and decoding phase are now done separately.
Decoding phase will leverage KV cache and speed improves.

Requires a PyTorch side fix to work. Will change the pin version.

Requires pytorch PR https://github.com/pytorch/pytorch/pull/136243 to land.
Otherwise, you may hit issue described in: https://github.com/pytorch/pytorch/issues/136225

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -187,8 +187,8 @@ def _create_padded_prompts(
 
 def _batch_decode_next_tokens(
     output: torch.Tensor,
-    prompt_lengths: List[int],
     tokenizer,
+    prompt_lengths: Optional[List[int]] = None,
 ) -> List[Tuple[int, str]]:
     """
     Decode the next token for each prompt in the batch.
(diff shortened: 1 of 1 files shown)


dist_run.py
```diff
@@ -390,9 +394,10 @@ def main(args):
     padded_sequence, prompt_lengths = _create_padded_prompts(
         input_ids, tokenizer, seqlen, start_pos, device
     )
-
-    # create schedule
-    schedule = ScheduleGPipe(stage, mbs)
+    # TODO: figure out how to set input_pos for each prompt in the batch then we
```
lessw2020, line 397: Removing this limitation is probably our most important next step.

lessw2020 (approved):
Great work, looks great. Left comment but removing the all prompt same len is likely most important next step. Anyway this pr is a big step forward, nice job!

PR #1168, opened 2024-09-20
parallelize ExecuTorch build

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #1171
* #1170
* #1186
* #1184
* #1169
* __->__ #1168
* #1181

PR #1170, opened 2024-09-20
Default max_seq_length to 128 for ExecuTorch export

swolchok:
* #1171
* __->__ #1170
* #1186
* #1184
* #1169
* #1168
* #1181

With the current default behavior, performance for e.g. stories110Mwithout custom SDPA is bad because the QKV tensors are long (8192 in the last dim). Limiting the max sequence length remedies this.


Jack-Khuu (approved):

desertfire (approved):

swolchok:
looks like we need to fix torchchat not respecting max_seq_length from PTEs in order to land this

PR #1174, opened 2024-09-21
[Distributed] Add lanes to KV cache

kwen2501:
KV cache is extended to have multiple lanes, each letting a separate batch pass through, achieving pipeline parallelism.

    # The number of cache lanes is the same as the maximum number of
    # micro-batches that can be "in flight" in parallel -- imagine each
    # micro-batch takes 1 "pipeline lane," they need distinct KV cache spaces.
    # When decoding is done for certain micro-batches, we can reuse the KV cache
    # lanes.

### Major changes
1. `setup_caches` will take one kwarg `cache_lanes` (default to 1).
```
def setup_caches(self, max_batch_size, max_seq_length, cache_lanes: int = 1)
```

2. `attention.kv_cache` is now a `nn.ModuleList`, containing multiple `KVCache`'s, each corresponding to a lane.

3. We now pass `kwargs = {"input_pos": input_pos, "cache_lane": lane}` to the `step()` function. Removing the temporary helper function `model.setup_input_pos`.

Requires https://github.com/pytorch/pytorch/pull/136416 to support pass-in of kwargs.

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -359,13 +359,13 @@ def main(args):
         output_args=example_outputs,
         group=pp_group,
     )
+
+    # Create schedule
     # Number of micro-batches for the schedule is 1, because each step() call we
     # only push 1 micro-batch into the pipeline. But we can continuously push
     # new micro-batches into the pipeline as they arrive, achieving same
     # pipelining effect.
-    mbs = 1
-    # create schedule
-    prefiller = ScheduleGPipe(prefill_stage, mbs)
+    prefiller = ScheduleGPipe(prefill_stage, 1)
 
     prompt = [
         "What is a computer?",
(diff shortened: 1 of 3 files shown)


dist_run.py
```diff
@@ -445,11 +456,12 @@ def get_example_ins_outs(seqlen: int) -> Tuple[torch.Tensor, torch.Tensor]:
         group=pp_group,
     )
     # create schedule
-    decode_schedule = ScheduleGPipe(decode_stage, mbs)
+    decorder = ScheduleGPipe(decode_stage, 1)
```
lessw2020, line 459: syntax error - this should be 'decoder' and not 'decorder'.

dist_run.py
```diff
@@ -467,11 +479,11 @@ def get_example_ins_outs(seqlen: int) -> Tuple[torch.Tensor, torch.Tensor]:
 
             # Run data through pipeline
             if pp_rank == first_pp_rank:
-                output = decode_schedule.step(new_token)
+                output = decorder.step(new_token, **kwargs)
             elif pp_rank == last_pp_rank:
-                output = decode_schedule.step()
+                output = decorder.step(**kwargs)
```
lessw2020, line 484: ''

lessw2020 (approved):
nice addition!
minor note that 'decorder' should be 'decoder' in the code for ease of understanding/syntax.

PR #1180, opened 2024-09-23
[Distributed] Make decode in flight optional

kwen2501:
Decoding tensor-form token ids to strings involves a CPU sync.

This PR added a flag to disable on-the-flight conversion.
```
torchrun --nproc-per-node 4 dist_run.py llama2-7b-chat --pp 2 --disable-in-flight-decode
```

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -399,10 +399,9 @@ def main(args):
     last_pp_rank_global_id = dist.get_global_rank(pp_group, last_pp_rank)
 
     # New token generated each iteration
-    total_prompts = len(prompt_lengths)
-    # need a new token dimension (row) for each prompt in the batch
-    new_token = torch.zeros(total_prompts, 1, device=device, dtype=torch.int64)
-    res = [[] for _ in range(total_prompts)]
+    # need a row dimension for each prompt in the batch
+    new_token = torch.zeros(batch_size, 1, device=device, dtype=torch.int64)
+    res = [[] for _ in range(batch_size)]
     num_tokens = 40
 
     # Prefill phase
--- a/dist_run.py
+++ b/dist_run.py
@@ -501,7 +501,8 @@ def main(args):
         res_list = res.tolist()
         response = tokenizer.decode(res_list)
         for i in range(len(response)):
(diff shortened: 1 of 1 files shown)


dist_run.py
```diff
@@ -399,11 +392,10 @@ def get_example_ins_outs(seqlen: int) -> Tuple[torch.Tensor, torch.Tensor]:
     last_pp_rank_global_id = dist.get_global_rank(pp_group, last_pp_rank)
 
     # New token generated each iteration
-    total_prompts = len(prompt_lengths)
-    # need a new token dimension (row) for each prompt in the batch
-    new_token = torch.zeros(total_prompts, 1, device=device, dtype=torch.int64)
-    res = [[] for _ in range(total_prompts)]
-    num_tokens = 40
+    # need a row dimension for each prompt in the batch
+    new_token = torch.zeros(batch_size, 1, device=device, dtype=torch.int64)
```
lessw2020, line 396: general comment- do we really need int64 for the dtype given that int32 holds to positive 2B? (meant to ask this earlier).
Seems like some minor savings and I don't see any vocabs getting to 2B anytime soon? 
Can just leave as is for consistency and maybe a single PR in the future to sweep this up if we agree int32 is fine.

lessw2020 (approved):
looks good!
longer term, I already started a PR similar to this one but with the idea of doing a chunky decoding, where there is a variable that represents how many tokens to generate between decoding.
The idea here is that we don't want to wait until the *entire* generation is done before showing things to the user.  But as this PR points out, we have a sync every time we decode...so maybe generating 20 tokens and then decode display, another round of generate/display, etc. so that we balance the tension between decoding speed and still updating the user as to the response. 
Anyway, we can discuss and can build on this option in this PR to make it more tunable. 
Thanks for adding this!

kwen2501:
Ah, sorry, didn't mean to "overwrite" your PR. I was just doing code cleaning and all of a sudden touched the decode path..

The chunk decode idea definitely sounds good! My flag of "disable-in-flight-decode" can be turned into chunk size = -1 (infinity) in your case.

PR #1182, opened 2024-09-24
Add torchao

metascroy:


runner/et.cmake
```diff
@@ -116,6 +116,13 @@ if(executorch_FOUND)
     target_link_libraries(et_run PRIVATE log)
   endif()
 
+  if(LINK_TORCHAO_OPS)
+    target_link_libraries(et_run PRIVATE "$<LINK_LIBRARY:WHOLE_ARCHIVE,${TORCHCHAT_ROOT}/torchao-build/cmake-out/lib/libtorchao_ops_executorch.a>")
```
metascroy, line 120: Change this line to:

```
target_link_libraries(et_run PRIVATE "${TORCHCHAT_ROOT}/torchao-build/cmake-out/lib/libtorchao_ops_EXECUTORCH.dylib")
```

.github/workflows/pull.yml
```diff
@@ -621,71 +621,87 @@ jobs:
           python torchchat.py remove stories15m
 
   test-mps:
-    uses: pytorch/test-infra/.github/workflows/macos_job.yml@main
```
metascroy, line 624: This test is very flaky because it fails whenever a machine with python 3.9 is allocated.  Change to always setup python 3.10.

.github/workflows/pull.yml
```diff
@@ -734,66 +750,82 @@ jobs:
 
           echo "Tests complete."
   test-mps-dtype:
-    uses: pytorch/test-infra/.github/workflows/macos_job.yml@main
```
metascroy, line 737: This test is very flaky because it fails whenever a machine with python 3.9 is allocated.  Change to always setup python 3.10.

torchchat/utils/scripts/install_et.sh
```diff
@@ -19,10 +19,4 @@ pushd ${TORCHCHAT_ROOT}
 find_cmake_prefix_path
 clone_executorch
 install_executorch_libs $ENABLE_ET_PYBIND
-install_executorch_python_libs $ENABLE_ET_PYBIND
```
metascroy, line 22: Move this to install_exeuctorch_python_libs because on the non-cache path in the CI, having it here doesn't work.

.github/workflows/pull.yml
Jack-Khuu: Thanks for fixing this test- it's been a thorn on our side for a while

docs/quantization.md
```diff
@@ -118,6 +118,75 @@ python3 torchchat.py export llama3 --quantize '{"embedding": {"bitwidth": 4, "gr
 python3 torchchat.py generate llama3 --pte-path llama3.pte  --prompt "Hello my name is"
 ```
 
+## Experimental TorchAO lowbit kernels
```
Jack-Khuu, line 121: I'll take pass at adding rich elements to this in a separate PR; Content looks good here

torchchat/utils/scripts/install_et.sh
Jack-Khuu: If we remove this line: `install_executorch_python_libs $ENABLE_ET_PYBIND`

Where is this being called?

Jack-Khuu (approved):

Jack-Khuu:
Can you also update the title and description?

torchchat/utils/scripts/install_et.sh
metascroy: It is called inside install_executorch_libs (which installs both c++ and python), so it was redundant

PR #1190, opened 2024-09-24
Use default max_seq_length of 128 when loading ExecuTorch models

swolchok:
We just defaulted on export, we should also default on load. ExecuTorch has API to get the max seq length, but the Python API currently segfaults when the method is missing.


swolchok:
* (to be filled)

Jack-Khuu (approved):

PR #1192, opened 2024-09-24
Default max_seq_length to 128 for ExecuTorch export

swolchok:
With the current default behavior, performance for e.g. stories110Mwithout custom SDPA is bad because the QKV tensors are long (8192 in the last dim). Limiting the max sequence length remedies this.


swolchok:
* (to be filled)

Jack-Khuu (approved):

PR #1194, opened 2024-09-24
[Distributed] Fix cache lane

kwen2501:
Previous PR https://github.com/pytorch/torchchat/pull/1174 has two issues:

(i) Wrong default `cache_lane` value given to forward function (should be 0, gave 1).

(ii) Did not add cache lane impl to `CustomKVCache`. The impl is required if we want the KVCache to CustomKVCache replacement work smoothly.


lessw2020 (approved):
thanks for the update - lgtm.

PR #1198, opened 2024-09-24
native runner: convert logits to float

swolchok:
The native runner incorrectly assumed that the output of the model
must always be float. Now it doesn't.

### DIFF
--- a/runner/run.cpp
+++ b/runner/run.cpp
@@ -212,6 +212,7 @@ float* forward(Transformer* transformer, int token, int pos) {
                              .to(torch::dtype(torch::kFloat32))
                              .to(torch::kCPU);
   auto logits = result[0].data_ptr();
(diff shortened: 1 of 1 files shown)


swolchok:
* (to be filled)

larryliu0820 (approved):

runner/run.cpp
```diff
@@ -212,6 +212,7 @@ float* forward(Transformer* transformer, int token, int pos) {
                              .to(torch::dtype(torch::kFloat32))
                              .to(torch::kCPU);
   auto logits = result[0].data_ptr();
+  memcpy(s->logits, logits, p->vocab_size * sizeof(float));
```
larryliu0820, line 215: This is for AOTI right? This is needed because the rest of the runner assumes float?

swolchok: it is simply moved up from below.

PR #1202, opened 2024-09-25
Update Flamingo Builder to use Llama3ScaledRoPE instead of RotaryPositionalEmbeddings

Gasoonjia:
We use a new rope class after bumped tune version.

### DIFF
--- a/torchchat/cli/builder.py
+++ b/torchchat/cli/builder.py
@@ -32,6 +32,7 @@ from torchtune.training import set_default_dtype
 from torchchat.model import Model, ModelArgs, ModelType
 
 from torchtune.modules.position_embeddings import RotaryPositionalEmbeddings
+from torchtune.models.llama3_1._position_embeddings import Llama3ScaledRoPE
 
 from torchchat.model_config.model_config import resolve_model_config
 from torchchat.utils.build_utils import (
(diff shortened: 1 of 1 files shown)


torchchat/cli/builder.py
```diff
@@ -32,6 +32,7 @@
 from torchchat.model import Model, ModelArgs, ModelType
 
 from torchtune.modules.position_embeddings import RotaryPositionalEmbeddings
```
Jack-Khuu, line 34: Let's get rid of the old one

Jack-Khuu (approved):

torchchat/cli/builder.py
Gasoonjia: done.

torchchat/cli/builder.py
```diff
@@ -402,7 +402,7 @@ def _load_model_default(builder_args: BuilderArgs) -> Model:
             max_seq_len = decoder_config['max_seq_len']
             rope_base = decoder_config['rope_base']
             for submodule in model.modules():
-                if isinstance(submodule, RotaryPositionalEmbeddings):
+                if isinstance(submodule, Llama3ScaledRoPE):
                     submodule.__init__(head_dim, max_seq_len, rope_base)
```
Jack-Khuu, line 406: nit: We should we populating the rest of the RoPE fields based on the config

From tune:
```
scale_factor: int = 8,
low_freq_factor: int = 1,
high_freq_factor: int = 4,
old_context_len: int = 8192,
```

PR #1203, opened 2024-09-25
[Distributed] Move into torchchat folder

kwen2501:
Moved `distributed` folder into `torchchat` folder.

`dist_run.py` unfortunately have to stay at root a bit longer till either of the following happens:
(i) we support `pip install` of torchchat (needs a `pyproject.toml`) or
(ii) we refactor `dist_run.py` as a "main" backend callable by `torchchat.py`, like:
https://github.com/pytorch/torchchat/blob/0b8ca05d0a001f7f858ced6c5b8839c076eae72a/torchchat.py#L79-L92

Item (ii) should be on our agenda.


run_dist.sh
```diff
@@ -1,7 +0,0 @@
-#export CUDA_VISIBLE_DEVICES=4,5,6,7
-PORT=${1:-29501}
-NGPU=${NGPU:-"4"}
-LOG_RANK=${LOG_RANK:-0,1,2,3}
```
lessw2020, line 4: ouch...lol.

lessw2020: I'd prefer to leave this file in for now.

PR #1204, opened 2024-09-25
🦙🦙 Llama3.2 Release: Support for 1B, 3B, and 11B (Multimodal)

Jack-Khuu:
This PR introduces support for the new Llama3.2 models announced at Connect 2024.

**Llama3.2 1B and 3B** are text-only models that now have access to all the same torchchat features as their predecesessors (Llama3.1 8B)

**Llama3.2 11B** is a Multimodal Model Capable of supporting both Image and Text inputs. For more information about what is currently supported, take a look at https://github.com/pytorch/torchchat/blob/llama32/docs/multimodal.md


larryliu0820 (approved):

iseeyuan (approved):
Nice!

PR #1209, opened 2024-09-25
Move back to nightly for torchtune instead of commithash

Jack-Khuu:


Jack-Khuu:
Dropping thing for now until we get the CI back up

Will spin up again later

PR #1212, opened 2024-09-26
Make default max_seq_length in openai_api consistent to chat

iseeyuan:
Those two apis should have the same default max_seq_length. There can be a better location to define this. The max_seq_length in chat is 2048, in https://github.com/pytorch/torchchat/blob/main/torchchat/generate.py#L821.

The way max_seq_length defined in model looks flaky. It should be consistent to the model definition and the cache setup.


Jack-Khuu:
Absorbed in https://github.com/pytorch/torchchat/pull/1211/files

PR #1215, opened 2024-09-26
Fix eval sanity check CI

larryliu0820:
Summary: Unbreak `test-cpu-eval-sanity-check-*` CI jobs

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -905,13 +905,11 @@ jobs:
       - name: Load or install ET
         id: install-et
         uses: actions/cache@v4
-        env:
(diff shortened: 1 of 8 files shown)


byjlw (approved):
do we still need the big machine here?

larryliu0820:
> do we still need the big machine here?

Ideally all CI jobs should use PyTorch test infra workflows, meaning they should have this line:

```
    uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
```

It provides ssh functionality automatically and probably other benefits. However it's not critical to fix CI so reverting it

Jack-Khuu (approved):

Jack-Khuu:
Landing this PR as-is to unblock + jobs stuck in queue

Will have separate PR for any other fixes

PR #1216, opened 2024-09-26
support text-only input with llama3.2-11b

Gasoonjia:
This PR enables llama3.2-11b model with text-only input.
Note that this PR is only for CLI pipeline. Will have a separate PR for openai-api update

```
(torchchat-test) [gasoonjia@server ~/torchchat-32mm (main|REBASE-i|main)]$ python torchchat.py generate llama3.2-11B --prompt "How are you these days"
Using device=cuda NVIDIA PG509-210
Loading model...
Time to load model: 8.88 seconds
How are you these daysI'm just a computer program, so I don't have feelings or emotions like humans do, but thanks for asking! I'm functioning properly and ready to assist with any questions or tasks you may have. How about you? How's your day going?


      Average tokens/sec (total): 11.66                 
Average tokens/sec (first token): 1.82                 
Average tokens/sec (next tokens): 13.04 
```

### DIFF
--- a/torchchat/generate.py
+++ b/torchchat/generate.py
@@ -353,27 +353,34 @@ class Generator:
         width = x.size(1)
         assert input_pos.size(0) == width
 
-        if batch is not None:
+        if self.model.config.model_type == ModelType.Flamingo:
+            assert batch is not None, "Flamingo requires batch"
+
             # TODO: Verify sequential prefill works with multimodal models
-            tokens = batch["tokens"]
+            is_multimodal = True
             if 'encoder_input' in batch:
                 encoder_input = batch['encoder_input']
+                encoder_mask = batch["encoder_mask"]
+                is_multimodal = True
             else:
                 encoder_input = None
+                encoder_mask = None
+                is_multimodal = False
 
-            seq_len = tokens.size(1)
+            seq_len = x.size(1)
             mask = batch["causal_mask"][None, :seq_len]
-            encoder_mask = batch["encoder_mask"]
             input_pos = input_pos.view(1, -1)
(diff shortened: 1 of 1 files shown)


torchchat/generate.py
```diff
@@ -353,27 +353,34 @@ def prefill(
...
                 encoder_input = batch['encoder_input']
+                encoder_mask = batch["encoder_mask"]
+                is_multimodal = True
             else:
                 encoder_input = None
+                encoder_mask = None
+                is_multimodal = False
 
-            seq_len = tokens.size(1)
+            seq_len = x.size(1)
```
iseeyuan, line 370: What's the difference between tokens and x?

iseeyuan (approved):

joecummings:
You can also check for multimodal content like this so you don't have to check the model type explicitly:

https://github.com/pytorch/torchtune/blob/7da96d18dbddbfcf77e045fe149b1efb6866681d/recipes/dev/generate_v2.py#L123

torchchat/generate.py
Gasoonjia: No difference, all equal to text input
Previously we pull the text input out of `batch` inside prefill and call it `tokens`
Now we pull out text input first then forward it into prefill function and call it `x`
The reason we change the name is other single modality models usually use `x` to represent the text token inputs.

Gasoonjia:
@joecummings thanks for your suggestions!
Just curious how do you guys handle the mismatch between input and model requirements? Like when user forward an image input into a single-modality model?

Jack-Khuu:
The failing test is not relevant. Pushing through

PR #1225, opened 2024-09-28
[Distributed] Bump torch version

kwen2501:
Distributed inference (pipeline parallel part, to be specific) requires two features landed in pytorch nightly:

- [[Pipelining] Make PipelineStage support meta initialization](https://github.com/pytorch/pytorch/pull/136243)

- [[Pipelining] Allow non-0 stages to accept kwargs](https://github.com/pytorch/pytorch/pull/136416)

Thus bumping the torch version to `2.6.0.dev20240925`.

Tested locally.

Cc: @Jack-Khuu @lessw2020


install/install_requirements.sh
```diff
@@ -47,7 +47,7 @@ fi
 # NOTE: If a newly-fetched version of the executorch repo changes the value of
 # PYTORCH_NIGHTLY_VERSION, you should re-run this script to install the necessary
 # package versions.
-PYTORCH_NIGHTLY_VERSION=dev20240901
+PYTORCH_NIGHTLY_VERSION=dev20240925
 
 # Nightly version for torchvision
 VISION_NIGHTLY_VERSION=dev20240901
```
lessw2020, line 53: we need to bump the vision version.  
atm I get this error running this PR:
ERROR: Cannot install torch==2.6.0.dev20240925 and torchvision==0.20.0.dev20240901+cu121 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested torch==2.6.0.dev20240925
    torchvision 0.20.0.dev20240901+cu121 depends on torch==2.5.0.dev20240901

lessw2020 (commented):
we need to bump the nightly vision version to avoid an error. 
i.e. this: 
VISION_NIGHTLY_VERSION=dev20240901

Bumping to same as PYTORCH_NIGHTLY_VERSION fixes this error for me.
See inline comment.

Jack-Khuu:
Vision and PT being on the same nightly pin usually solves this

Jack-Khuu:
I'll also be bumping the version over on ExecuTorch https://github.com/pytorch/executorch/pull/5549

I can match you at 925

kwen2501:
@metascroy Wondering if you could have a look at the CI issue? Thanks!

metascroy:
> @metascroy Wondering if you could have a look at the CI issue? Thanks!

At an initial glance, I'm not sure why it is failing.  I'll investigate more after lunch, but I'm wondering if PyTorch changed how custom ops work in the pin bump.

kwen2501:
Cc @zou3519 any recent changes in custom op support that may be relevant to the CI failure here?

zou3519:
Maybe @jerryzh168 ?

metascroy:
> > @metascroy Wondering if you could have a look at the CI issue? Thanks!
> 
> At an initial glance, I'm not sure why it is failing. I'll investigate more after lunch, but I'm wondering if PyTorch changed how custom ops work in the pin bump.

@kwen2501 I figured out the issue.

ExecuTorch uses PyTorch 9/1, so when ExecuTorch installs, it installs 9/1 and the two different PyTorch versions cause an issue.  In https://github.com/pytorch/torchchat/pull/1235/, I bump the PT pin, but remove the ET parts of the test and it passes.

So before PT pin can be updated, it needs to be updated in ET first and the PT/ET pin then need to be updated together in torchchat cc @Jack-Khuu

kwen2501:
Thanks @metascroy !

> I'll also be bumping the version over on ExecuTorch [pytorch/executorch#5549](https://github.com/pytorch/executorch/pull/5549)
> 
> I can match you at 925

@Jack-Khuu Looks like this bump here needs to wait for your bump in ET to land first :)

Jack-Khuu:
Unified pin bump https://github.com/pytorch/torchchat/pull/1269

PR #1227, opened 2024-09-28
[Distributed] create model on meta device

kwen2501:
### Change 1
Changing from:
```
    with device:
        model = Transformer(config)
```
to
```
    with torch.device("meta"):
        model = Transformer(config)
```
because when we later load weights, we will swap out the tensors of the model anyway. So saving an on-device init here.

### Change 2
Also added a `with device` context to `model.setup_cache()` call. So that the caches are directly created on the target device -- saving a `model.to(device)` call.

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -328,14 +328,25 @@ def main(args):
     config.stage_idx = pp_rank
     config.n_stages = pp_degree
 
-    with device:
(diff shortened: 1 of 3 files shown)


lessw2020 (approved):
lgtm!

Issue #1220, opened 2024-09-27
lm_eval not working in the model.py


HamidShojanazeri:
### 🐛 Describe the bug

Following the installation README, and running dist_run.py, it fails to import utils from evaluate module. It also seems not being used atm, may as well remove it?

```
(torchchat)  hamidnazeri@submit-0:~/torchchat$ python dist_run.py --model_name meta-llama/Meta-Llama-3-8B-Instruct
Traceback (most recent call last):
  File "/opt/hpcaas/.mounts/fs-0301404b74c8d22fd/home/hamidnazeri/torchchat/dist_run.py", line 37, in <module>
    from torchchat.cli.builder import _initialize_tokenizer, TokenizerArgs
  File "/opt/hpcaas/.mounts/fs-0301404b74c8d22fd/home/hamidnazeri/torchchat/torchchat/cli/builder.py", line 29, in <module>
    from torchchat.model import Model, ModelArgs, ModelType
  File "/opt/hpcaas/.mounts/fs-0301404b74c8d22fd/home/hamidnazeri/torchchat/torchchat/model.py", line 35, in <module>
    import lm_eval  # noqa 
  File "/data/home/hamidnazeri/miniconda3-2/envs/torchchat/lib/python3.10/site-packages/lm_eval/__init__.py", line 1, in <module>
    from .evaluator import evaluate, simple_evaluate
  File "/data/home/hamidnazeri/miniconda3-2/envs/torchchat/lib/python3.10/site-packages/lm_eval/evaluator.py", line 11, in <module>
    import lm_eval.api.metrics
  File "/data/home/hamidnazeri/miniconda3-2/envs/torchchat/lib/python3.10/site-packages/lm_eval/api/metrics.py", line 168, in <module>
    exact_match = hf_evaluate.load("exact_match")
  File "/data/home/hamidnazeri/miniconda3-2/envs/torchchat/lib/python3.10/site-packages/evaluate/loading.py", line 751, in load
    evaluation_cls = import_main_class(evaluation_module.module_path)
  File "/data/home/hamidnazeri/miniconda3-2/envs/torchchat/lib/python3.10/site-packages/evaluate/loading.py", line 76, in import_main_class
    module = importlib.import_module(module_path)
  File "/data/home/hamidnazeri/miniconda3-2/envs/torchchat/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
[...]

PR #1228, opened 2024-09-28
Wrap unused lm_eval in try-catch: Unblock dist_run

Jack-Khuu:
The existing code results in an unhealthy import environment when executing from a distributed setting i.e. `dist_run.py`. This is continuation of the fix from https://github.com/pytorch/torchchat/commit/ec7b51070d8b1f950bf67b133f057822fe53240b

The try-catch is a temporary patch to unbreak dist_run.py, while we investigate. Bug: https://github.com/pytorch/torchchat/issues/1220

---
Test:
`python dist_run.py --model_name` no longer fails ion import (see bug issue for details)

### DIFF
--- a/install/install_requirements.sh
+++ b/install/install_requirements.sh
@@ -52,9 +52,6 @@ PYTORCH_NIGHTLY_VERSION=dev20240901
 # Nightly version for torchvision
 VISION_NIGHTLY_VERSION=dev20240901
 
-# Nightly version for torchtune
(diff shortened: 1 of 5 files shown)


kwen2501 (approved):
Thanks much for fixing this!!

kwen2501:
Resolves #1220 ?

Jack-Khuu:
> Resolves #1220 ?

That's what I'm hoping for haha

PR #1234, opened 2024-09-30
Implement universal batch_decode & decode_in_flight for llama2 & llama3, with deterministic or multinomial (topk) decoding (handle both sentencepiece (llama2) and tiktoken (llama3))

lessw2020:
This PR:
1 - updates the batch_decode_next_tokens in a way that handles both llama2 and llama3 with their respective tokenizers.  Thus we have a single universal decoding for in flight. 

2 - adds a temperature option to enable non-deterministic (creative) decoding, using topk and multinomial selection.

3 - update decode_in_flight, again to be compat with both llama2 and llama3. 

4 - minor tweaks to using zip for final display and move decode_in_flight to _decode_in_flight with other global functions for ease of reference.

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -16,7 +16,6 @@ from typing import Any, Dict, List, Optional, Tuple
 
 import torch
 import torch.distributed as dist
-import torch.nn.functional as F
 from torch.distributed.pipelining import PipelineStage, ScheduleGPipe
 from torchchat.cli.builder import _initialize_tokenizer, TokenizerArgs
 
--- a/dist_run.py
+++ b/dist_run.py
@@ -375,7 +375,7 @@ def main(args):
 
     prompt = [
         "What is Snow?",
-        "What is Fire?",
+        "Who is Santa Claus?",
         # "Where does Santa live?",
         # "Who is Abraham Lincoln?",
         # "How are models trained?",
--- a/dist_run.py
+++ b/dist_run.py
@@ -16,6 +16,7 @@ from typing import Any, Dict, List, Optional, Tuple
 
 import torch
 import torch.distributed as dist
+import torch.nn.functional as F
 from torch.distributed.pipelining import PipelineStage, ScheduleGPipe
 from torchchat.cli.builder import _initialize_tokenizer, TokenizerArgs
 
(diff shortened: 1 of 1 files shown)


metascroy:
Rebase to fix failing tochao_experimental check

dist_run.py
```diff
@@ -416,23 +455,13 @@ def get_example_ins_outs(seqlen: int) -> Tuple[torch.Tensor, torch.Tensor]:
...
-
     # Decode the output -- first generated token
     if pp_rank == last_pp_rank:
-        new_token = _batch_decode_next_tokens(output, prompt_lengths[0] - 1)
+        logger.info(f"{color.green}Decoding...{prompt_lengths=}{color.reset}")
+        new_token = _batch_decode_next_tokens(output, prompt_lengths)
         res.append(new_token)
         if not args.disable_in_flight_decode:
-            decode_in_flight(new_token)
+            _decode_in_flight(new_token, tokenizer, tp_rank)
```
kwen2501, line 464: nit: put `tp_rank` to the `if` condition?

dist_run.py
```diff
@@ -482,10 +511,11 @@ def decode_in_flight(token):
 
             # Decode the output
             if pp_rank == last_pp_rank:
-                new_token = _batch_decode_next_tokens(output, 0)
+                # logger.info(f"{color.red}Decoding...{output.shape=}{color.reset}")
```
kwen2501, line 514: nit: remove log?

dist_run.py
```diff
@@ -499,12 +529,17 @@ def decode_in_flight(token):
...
-        response = tokenizer.decode(res_list)
-        for i in range(len(response)):
-            logger.info(f"Prompt: {color.green}{prompt[i]} {color.reset}")
-            logger.info(f"Response: {color.red}{response[i]} {color.reset}")
+
+        res_stacked = torch.stack(res, dim=1)
+        res_list = res_stacked.tolist()
+
+        # Decode the output as comprehension instead of loop
+        responses = [tokenizer.decode(sequence) for sequence in res_list]
```
kwen2501, line 537: Hmm, did the previous code not work in case of variable length? Just curious.
`response = tokenizer.decode(res_list)`

dist_run.py
```diff
@@ -189,23 +189,49 @@ def _create_padded_prompts(
...
-    next_token = torch.argmax(next_token_logits, dim=-1)
-    # Token ids in int tensor form
-    return next_token
+    batch_size, seq_len, vocab_size = output.shape
+
+    if step != -1:
+        next_token_logits = output[:, 0, :]
+    else:
+        # get the logits for each prompt at the specified positions
+        next_token_logits = output[torch.arange(batch_size), torch.tensor(pos) - 1]
```
kwen2501, line 215: Hmm, why "-1"?
From this function's perspective, if the caller has given the position, should it just faithfully decode that position?
(I understand that this can be run right if providing prompt_length instead of promt_length -1 at callsite.)

kwen2501 (approved):
Thanks for the new feature! LGTM. Just some minor comments.

lessw2020:
* will revisit to update re: comments.  Landing for now so we can work on tp fix.

PR #1235, opened 2024-09-30
[Not for landing] Update PT pin

metascroy:


lessw2020:
we should close out this related PR if this one lands as they are the same thing (just marked under [Distributed]:
https://github.com/pytorch/torchchat/pull/1225/files

metascroy:
> we should close out this related PR if this one lands as they are the same thing (just marked under [Distributed]: https://github.com/pytorch/torchchat/pull/1225/files

This one is not for landing.  It was to test hypothesis that pin bump error had to due with ET conflict.

PR #1247, opened 2024-10-01
Fix non-MM multiturn: Use legacy formatting

Jack-Khuu:
Refactor introduced in https://github.com/pytorch/torchchat/commit/8278aa26fff3ac0162b7eaafe4dc41445f94d867 prevented non multimodal models (e.g. 3.1 8B) from doing multiturn in borswer/openai api.

This re-enables it via casing out MM (LLama3.2 11B). Support for 11B Multiturn and reunification of logic will be in followup PRS

---
Tested via Browser
```
python torchchat.py server llama3.2-11B 
python torchchat.py server llama3.2-1B 
```

Then test in browser with 1B (Text only - Saw it can do multiturn) and 11B (Multimodal - No multiturn as expected)
```
 streamlit run torchchat/usages/browser.py
```

### DIFF
--- a/torchchat/usages/openai_api.py
+++ b/torchchat/usages/openai_api.py
@@ -21,6 +21,7 @@ from PIL import Image
 
 from torchchat.cli.download import is_model_downloaded, load_model_configs
 from torchchat.generate import Generator, GeneratorArgs
+from torchchat.model import FlamingoModel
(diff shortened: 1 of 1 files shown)


vmpuri (approved):
lgtm

PR #1255, opened 2024-10-02
Safetensors

### LINKED ISSUE #1249: Support Huggingface models from safetensors

There are many models on Huggingface that are published as `safetensors` rather than `model.pth` checkpoints. The request here is to support converting and loading those checkpoints into a format that is usable with `torchchat`.

There are several places where this limitation is currently enforced:

* [_download_hf_snapshot](https://github.com/pytorch/torchchat/blob/main/torchchat/cli/download.py#L36) method explicitly ignores `safetensors` files.
* [convert_hf_checkpoint](https://github.com/pytorch/torchchat/blob/main/torchchat/cli/convert_hf_checkpoint.py#L44) explicitly looks for `pytorch_model.bin.index.json` which would be named differently for models that use `safetensors` (e.g. `model.safetensors.index.json`)
* [convert_hf_checkpoint](https://github.com/pytorch/torchchat/blob/main/torchchat/cli/convert_hf_checkpoint.py#L99) only supports `torch.load` to load the `state_dict` rather than `safetensors.torch.load`

### Alternatives

Currently, this `safetensors` -> `model.pth` can be accomplished manually after downloading a model locally, so this could be solved with documentation instead of code.

### Additional context

This issue is a piece of the puzzle for adding support for Granite Code 3b/8b which use the `llama` architecture in `transormers`, but take advantage several pieces of the architecture that are not currently supported by `torchchat`. The work-in-progress for Granite Code can be found on my fork: https://github.com/gabe-l-hart/torchchat/tree/GraniteCodeSupport

### RFC (Optional)

I have a working implementation to support `safetensors` during download and conversion that I plan to submit as a PR. The changes address the three points in code referenced above:

1. Allow the download of `safetensors` files in `_download_hf_snapshot`
    * I'm not yet sure how to avoid double-downloading weights for models that have both `safetensors` and `model.pth`, so will look to solve this before concluding the work
2. When looking for the tensor index file, search for all files ending in `.index.json`, and if a single file is found, use that one
3. When loading the `state_dict`, use the correct method based on the type of file (`torch.load` or `safetensors.torch.load`)

gabe-l-hart:
## Description

Closes #1249

This PR implements support for downloading and converting model checkpoints from `huggingface` which use the `safetensors` format rather than the `pth` binary (pickle) format.

## Changes

* Allow the tensor map file to be found under different `*.index.json` names
* Allow loading model files with `safetensors.torch.load` when needed
* Allow downloading safetensor files if they exist without pth files or the model explicitly prefers them

## Testing

* I have tested that the download and load for `llama3.1` are un-changed with these changes (no `safetensors` are downloaded)
* I have verified (on my [WIP branch for Granite Code](https://github.com/gabe-l-hart/torchchat/tree/GraniteCodeSupport
)) that for models with only `safetensors`, they are not ignored and can be cleanly converted

### DIFF
--- a/torchchat/cli/convert_hf_checkpoint.py
+++ b/torchchat/cli/convert_hf_checkpoint.py
@@ -102,9 +102,33 @@ def convert_hf_checkpoint(
 
     merged_result = {}
     for file in sorted(bin_files):
-        state_dict = torch.load(
+
+        # The state_dict can be loaded from either a torch zip file or
+        # safetensors. We take our best guess from the name and try all
+        # possibilities
+        load_pt_mmap = lambda: torch.load(
             str(file), map_location="cpu", mmap=True, weights_only=True
         )
+        load_pt_no_mmap = lambda: torch.load(
+            str(file), map_location="cpu", mmap=False, weights_only=True
+        )
+        def load_safetensors():
+            import safetensors.torch
+            with open(file, "rb") as handle:
+                return safetensors.torch.load(handle.read())
+        if "safetensors" in str(file):
+            loaders = [load_safetensors, load_pt_mmap, load_pt_no_mmap]
+        else:
+            loaders = [load_pt_mmap, load_pt_no_mmap, load_safetensors]
+
+        state_dict = None
+        for loader in loaders:
+            try:
+                state_dict = loader()
+                break
+            except Exception:
+                continue
+        assert state_dict is not None, f"Unable to load tensors from {file}"
         merged_result.update(state_dict)
     final_result = {}
     for key, value in merged_result.items():
--- a/torchchat/cli/convert_hf_checkpoint.py
+++ b/torchchat/cli/convert_hf_checkpoint.py
@@ -3,6 +3,7 @@
 
 # This source code is licensed under the license found in the
 # LICENSE file in the root directory of this source tree.
+import glob
 import json
 import os
 import re
@@ -41,7 +42,12 @@ def convert_hf_checkpoint(
     print(f"Model config {config.__dict__}")
 
     # Load the json file containing weight mapping
-    model_map_json = model_dir / "pytorch_model.bin.index.json"
+    model_map_json_matches = [Path(m) for m in glob.glob(str(model_dir / "*.index.json"))]
+    assert len(model_map_json_matches) <= 1, "Found multiple weight mapping files"
+    if len(model_map_json_matches):
+        model_map_json = model_map_json_matches[0]
+    else:
+        model_map_json = model_dir / "pytorch_model.bin.index.json"
 
     # If there is no weight mapping, check for a consolidated model and
     # tokenizer we can move. Llama 2 and Mistral have weight mappings, while
--- a/torchchat/cli/download.py
+++ b/torchchat/cli/download.py
@@ -22,18 +22,44 @@ from torchchat.model_config.model_config import (
 def _download_hf_snapshot(
     model_config: ModelConfig, artifact_dir: Path, hf_token: Optional[str]
 ):
-    from huggingface_hub import snapshot_download
+    from huggingface_hub import model_info, snapshot_download
     from requests.exceptions import HTTPError
 
     # Download and store the HF model artifacts.
     print(f"Downloading {model_config.name} from HuggingFace...", file=sys.stderr)
     try:
+        # Fetch the info about the model's repo
+        model_info = model_info(model_config.distribution_path, token=hf_token)
+        model_fnames = [f.rfilename for f in model_info.siblings]
+
+        # Check the model config for preference between safetensors and pth
+        has_pth = any(f.endswith(".pth") for f in model_fnames)
+        has_safetensors = any(f.endswith(".safetensors") for f in model_fnames)
+
+        # If told to prefer safetensors, ignore pth files
+        if model_config.prefer_safetensors:
+            if not has_safetensors:
+                print(
+                    f"Model {model_config.name} does not have safetensors files, but prefer_safetensors is set to True. Using pth files instead.",
+                    file=sys.stderr,
+                )
+                exit(1)
+            ignore_patterns = "*.pth"
+
+        # If the model has both, prefer pth files over safetensors
+        elif has_pth and has_safetensors:
+            ignore_patterns = "*safetensors*"
+
+        # Otherwise, download everything
+        else:
+            ignore_patterns = None
+
         snapshot_download(
             model_config.distribution_path,
             local_dir=artifact_dir,
             local_dir_use_symlinks=False,
             token=hf_token,
-            ignore_patterns="*safetensors*",
+            ignore_patterns=ignore_patterns,
         )
     except HTTPError as e:
         if e.response.status_code == 401:  # Missing HuggingFace CLI login.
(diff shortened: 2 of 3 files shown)


gabe-l-hart:
NOTE: The CLA is in process since I'm contributing through the IBM corporate CLA.

byjlw:
Thanks @gabe-l-hart
I appreciate you contributing. Code looks good, please rebase when you're ready and i'll approve and merge. 
Also since it looks like you're already adding another model to the list and have tested it feel free to bring in the definition so others can benefit by downloading via the cli

gabe-l-hart:
Thanks @byjlw! I'll get the rebase done shortly. As you noted, this is part of my work to add Granite Code support. I have a single [branch pointer](https://github.com/gabe-l-hart/torchchat/tree/GraniteCodeSupport) for all that work, but I have also tried to organize the commits so that they can be merged as bite-sized features rather than a big overhaul. Right now, the addition of the configs for Granite Code are at the end of the chain since the model won't work at all until the other features are present (it also currently only works in the python runtime). I'm totally open to whatever convention the `torchchat` team prefers in terms of contribution chunking, so just let me know what you prefer!

byjlw:
chunks are definitely better. Would love to learn more about the overall goals for the Granite code support

gabe-l-hart:
>Would love to learn more about the overall goals for the Granite code support

Good point, I didn't ever spell this out! Let me open a top-level issue about that support that I can use to track it all.

gabe-l-hart:
Top-level Granite Code support issue: https://github.com/pytorch/torchchat/issues/1262

byjlw:
Actually though, when i tested it by changing the the model.json to use safetensors for 11b base download errored out. It looks like this code works as long as it's not using definitions that use the torchtune format. Do you mind testing this case and resolving the issue?

```
"meta-llama/Llama-3.2-11B-Vision": {
        "aliases": ["llama3.2-11B-base", "Llama-3.2-11B-Vision-base"],
        "distribution_channel": "HuggingFaceSnapshot",
        "distribution_path": "meta-llama/Llama-3.2-11B-Vision",
        "prefer_safetensors": true
    },
```

```
Fetching 19 files: 100%|█████████████████████████████████████████████████████████████████████| 19/19 [1:06:30<00:00, 210.04s/it]
Converting meta-llama/Llama-3.2-11B-Vision to torchtune format...
Traceback (most recent call last):100%|████████████████████████████████████████████████████| 4.99G/4.99G [30:11<00:00, 3.52MB/s]
  File "/Users/byjlw/Documents/source/working/torchchat/torchchat.py", line 85, in <module>| 4.92G/4.92G [12:17<00:00, 3.34MB/s]
    check_args(args, "generate")
  File "/Users/byjlw/Documents/source/working/torchchat/torchchat/cli/cli.py", line 52, in check_args
    download_and_convert(args.model, args.model_directory, args.hf_token)
  File "/Users/byjlw/Documents/source/working/torchchat/torchchat/cli/download.py", line 123, in download_and_convert
    _download_hf_snapshot(model_config, temp_dir, hf_token)
  File "/Users/byjlw/Documents/source/working/torchchat/torchchat/cli/download.py", line 82, in _download_hf_snapshot
    convert_hf_checkpoint_to_tune( model_dir=artifact_dir, model_name=model_config.name)
  File "/Users/byjlw/Documents/source/working/torchchat/.venv/lib/python3.11/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/byjlw/Documents/source/working/torchchat/torchchat/cli/convert_hf_checkpoint.py", line 184, in convert_hf_checkpoint_to_tune
    raise RuntimeError(f"Could not find {consolidated_pth}")
RuntimeError: Could not find /Users/byjlw/.torchchat/model-cache/downloads/meta-llama/Llama-3.2-11B-Vision/original/consolidated.pth

```

gabe-l-hart:
Ah, good catch! I guess `N == 2` is a small sample size. I'll dig into your error and see how far I can get.

byjlw:
@ebsmothers can also help

gabe-l-hart:
Great. I'll report progress or blockers as they come up. @ebsmothers let me know if you dig in and get anywhere!

ebsmothers:
@byjlw so I'm not super familiar with how torchchat is handling checkpoint conversion but if you're switching from the .pth format to .safetensors format you will no longer be able to just do the simple move that's happening [here](https://github.com/pytorch/torchchat/blob/34d68312005730a3e355c2a2257b282a4be1f711/torchchat/cli/convert_hf_checkpoint.py#L148-L149). The safetensors format for Llama 3.2 11B Vision distributes the model weights across multiple files (see [here](https://huggingface.co/meta-llama/Llama-3.2-11B-Vision/tree/main)) so they will need to be loaded and merged into a single state dict as we do in torchtune [here](https://github.com/pytorch/torchtune/blob/0a3762d058fd9d860606a3a6bbebf71ac593ab5c/torchtune/training/checkpointing/_checkpointer.py#L428-L440).

gabe-l-hart:
That makes sense. A similar approach is being taken in [convert_hf_checkpoint](https://github.com/pytorch/torchchat/blob/main/torchchat/cli/convert_hf_checkpoint.py#L44) to load the sharded weights and convert them into the `.pth` format needed by `torchchat`. If I'm understanding the logic in `convert_hf_checkpoint_to_tune` correctly, it looks like it's really just a simpler version of the logic in `convert_hf_checkpoint` that doesn't require the tensor renaming or permuting. If that's the case, I think the fix should be to hoist out the shard-loading logic into a helper and then only do the post-processing in `convert_hf_checkpoint` before resaving as `model.pth`.

I finally have the safetensors downloaded locally, so I'll see how far I can get with this approach.

gabe-l-hart:
I have the mechanics of this working for the `safetensors` weights with `Llama-3.2-11B-Vision-Instruct`, but it appears that there is a naming difference between the tensors in `original/consolidated.pth` and the `safetensors` weights. This is likely similar to the name remapping needed in the standard conversion logic.

Given that, and given the fact that these models _do_ already have checkpoints that match the target naming scheme, I think it might make sense to leave the PR as-is for now and not switch `prefer_safetensors` to `true` for these models. I could also see an argument for using `safetensors` all-the-way-through to avoid the known `pickle` vulnerabilities with `.pth`, but this PR doesn't address that anyway since the models are converted to `.pth` during the conversion process.

byjlw (approved):

torchchat/cli/convert_hf_checkpoint.py
```diff
@@ -41,7 +42,12 @@ def convert_hf_checkpoint(
     print(f"Model config {config.__dict__}")
 
     # Load the json file containing weight mapping
-    model_map_json = model_dir / "pytorch_model.bin.index.json"
+    model_map_json_matches = [Path(m) for m in glob.glob(str(model_dir / "*.index.json"))]
+    assert len(model_map_json_matches) <= 1, "Found multiple weight mapping files"
```
leseb, line 46: Why is this an error? Thanks!

gabe-l-hart: Good catch. Seem my response over on your PR: https://github.com/pytorch/torchchat/pull/1346#issuecomment-2460257435

PR #1257, opened 2024-10-03
[Distributed] Fix tiktokenizer decoding

kwen2501:
`TiktokenTokenizer` (used by llama3) and `SentencePieceProcessor` (used by llama2) seem to have different requirement on input shape (1D list or 2D list).

Adding a condition to separate the input prep for the two.

Also improved some logging.

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -442,7 +442,6 @@ def main(args):
     # New token generated each iteration
     # need a row dimension for each prompt in the batch
     new_token = torch.zeros(batch_size, 1, device=device, dtype=torch.int64)
-    logger.info(f"{color.green}{new_token.shape=}, {new_token=}{color.reset}")
     # Store the generated tokens
     res = []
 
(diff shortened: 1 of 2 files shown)


dist_run.py
```diff
@@ -541,7 +539,13 @@ def get_example_ins_outs(seqlen: int) -> Tuple[torch.Tensor, torch.Tensor]:
         # token ids. Thus cat'ing along dim 1.
         res = torch.cat(res, dim=1)
         res_list = res.tolist()
-        responses = tokenizer.decode(res_list)
+        if isinstance(tokenizer, TiktokenTokenizer):
```
lessw2020, line 542: nit - I think this type of check is better done directly in the _build_chat_tokenizer function and then we just have an enum for the tokenizer type and can correctly err out if a not recognized tokenizer. 
The reason for that is two fold:
a - we future proof ourself so that if llama4 has a new tokenizer we are not going back through the code trying to figure where all the 'tokenizer' type checks are, and updating. 
b - we only check once in a logical point, and then all other code uses the enum, and we have an upfront single failure point to err out if we are hitting an unrecognized tokenizer.  
As this code is currently, it assumes that if not tiktoken then it must be sentencepiece which is a brittle assumption long term.

kwen2501:
Thanks @lessw2020 . Do you know why TiktokenTokenizer cannot decode 2D list?

PR #1259, opened 2024-10-03
Bias tensors

### LINKED ISSUE #1250: Add support for separate bias tensors

In the `transformers` implementation of llama, there are optional `bias` tensors for the [LlamaMLP](https://github.com/pytorch/torchchat/issues/1041) and [LlamaAttention](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L299) modules. Several additional models (specifically Granite Code 3B and 8B) use the `llama` architecture and have these separate bias tensors.

The proposal here is to add the ability to indicate the presence of bias tensors in [TransformerArgs](https://github.com/pytorch/torchchat/blob/main/torchchat/model.py#L260) and then support loading them in [Attention](https://github.com/pytorch/torchchat/blob/main/torchchat/model.py#L726) and [FeedForward](https://github.com/pytorch/torchchat/blob/main/torchchat/model.py#L852)

### Alternatives

If this project is designed to be limited to official Llama models, these bias tensors are not needed.

### Additional context

This issue is a piece of the puzzle for adding support for Granite Code 3b/8b which use the llama architecture in transormers, but take advantage several pieces of the architecture that are not currently supported by torchchat. The work-in-progress for Granite Code can be found on my fork: https://github.com/gabe-l-hart/torchchat/tree/GraniteCodeSupport

### RFC (Optional)

I have a working implementation to support these optional bias tensors that I plan to submit as a PR. The changes are along the following lines:

* Add new parameters to `TransformerArgs` for attention and ffn bias
* Set the `bias` value based on these parameters in both the `Attention` and `FeedForward` modules
* Support mapping `.bias` tensor names in `convert_hf_checkpoint`
* Support permuting `.bias` tensors in `convert_hf_checkpoint`
* Support loading permuted `.bias` tensors in `model.py`

gabe-l-hart:
## Dependencies

- [x] Safetensors: #1255

## Issues

Closes #1250

## Description

This PR adds support for models which have `bias` tensors for the attention and ffn modules alongside the primary weight tensors.

## Changes

* Add the bias tensors to the `weight_map` in HF checkpoint conversion
* Handle merged `wqkv` tensors for bias as well as weights in HF checkpoint conversion
    * This includes changes to the permutation logic to support the shapes of the bias tensors. I leveraged the corresponding logic in [llama.cpp's converter](https://github.com/ggerganov/llama.cpp/blob/master/convert_hf_to_gguf.py#L1569).
* Add configs to `TransformerArgs` to allow models to indicate the presence of `attention_bias` and `feed_forward_bias` tensors
* Populate the `Attention` and `FeedForward` modules' tensors' `bias` arguments based on the config args

## Testing

In conjunction with my other changes for Granite Code, I've been able to validate that the results produced with this logic do produce the expected token sequence.

NOTE: If there's any preferred way to include unit tests along with the PR, please let me know and I can get them added! I don't see a familiar unit test structure in the project at this point, so I've been relying on local ad-hoc testing.

### DIFF
--- a/torchchat/cli/convert_hf_checkpoint.py
+++ b/torchchat/cli/convert_hf_checkpoint.py
@@ -81,10 +81,17 @@ def convert_hf_checkpoint(
         "model.layers.{}.self_attn.k_proj.weight": "layers.{}.attention.wk.weight",
         "model.layers.{}.self_attn.v_proj.weight": "layers.{}.attention.wv.weight",
         "model.layers.{}.self_attn.o_proj.weight": "layers.{}.attention.wo.weight",
+        "model.layers.{}.self_attn.q_proj.bias": "layers.{}.attention.wq.bias",
+        "model.layers.{}.self_attn.k_proj.bias": "layers.{}.attention.wk.bias",
+        "model.layers.{}.self_attn.v_proj.bias": "layers.{}.attention.wv.bias",
+        "model.layers.{}.self_attn.o_proj.bias": "layers.{}.attention.wo.bias",
         "model.layers.{}.self_attn.rotary_emb.inv_freq": None,
         "model.layers.{}.mlp.gate_proj.weight": "layers.{}.feed_forward.w1.weight",
         "model.layers.{}.mlp.up_proj.weight": "layers.{}.feed_forward.w3.weight",
         "model.layers.{}.mlp.down_proj.weight": "layers.{}.feed_forward.w2.weight",
+        "model.layers.{}.mlp.gate_proj.bias": "layers.{}.feed_forward.w1.bias",
+        "model.layers.{}.mlp.up_proj.bias": "layers.{}.feed_forward.w3.bias",
+        "model.layers.{}.mlp.down_proj.bias": "layers.{}.feed_forward.w2.bias",
         "model.layers.{}.input_layernorm.weight": "layers.{}.attention_norm.weight",
         "model.layers.{}.post_attention_layernorm.weight": "layers.{}.ffn_norm.weight",
         "model.norm.weight": "norm.weight",
@@ -135,17 +142,15 @@ def convert_hf_checkpoint(
         if "layers" in key:
             abstract_key = re.sub(r"(\d+)", "{}", key)
             layer_num = re.search(r"\d+", key).group(0)
-            new_key = weight_map[abstract_key]
-            if new_key is None:
-                continue
+            new_key = weight_map.get(abstract_key, abstract_key)
             new_key = new_key.format(layer_num)
         else:
-            new_key = weight_map[key]
+            new_key = weight_map.get(key, key)
 
         final_result[new_key] = value
 
     for key in tuple(final_result.keys()):
-        if "wq" in key:
+        if "wq.weight" in key:
             q = final_result[key]
             k = final_result[key.replace("wq", "wk")]
             v = final_result[key.replace("wq", "wv")]
--- a/torchchat/cli/convert_hf_checkpoint.py
+++ b/torchchat/cli/convert_hf_checkpoint.py
@@ -142,7 +142,9 @@ def convert_hf_checkpoint(
         if "layers" in key:
             abstract_key = re.sub(r"(\d+)", "{}", key)
             layer_num = re.search(r"\d+", key).group(0)
-            new_key = weight_map.get(abstract_key, abstract_key)
+            new_key = weight_map[abstract_key]
+            if new_key is None:
+                continue
             new_key = new_key.format(layer_num)
         else:
             new_key = weight_map[key]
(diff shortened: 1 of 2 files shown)


gabe-l-hart:
Thanks for the review/merge on #1255! This PR is now ready for review

mikekg:
Current tests are run through .github/workflows, with some scripts in .ci (including a script that can be used to ensure that code in  
 markdown files works).  

Or were you looking for "unit tests" of subcomponents with a Python driver?  If you are looking for python-level unit tests, I don't think we have any right now, but that doesn't mean we can't have any.  If you want to make a proposal, you might discuss with @byjlw @Jack-Khuu and @lessw2020.

gabe-l-hart:
Hi @mikekg thanks for the pointers! I was looking for unit tests, but I'll dig into the CI workflows too.

mikekg:
> Hi @mikekg thanks for the pointers! I was looking for unit tests, but I'll dig into the CI workflows too.

There aren't any unit tests today, the workflows run as shell scripts.  It's very conceivable to create a unit test directory and set up a workflow that runs them.  I think the most interesting question is what testing framework to use.  I don't have any string convictions about that, what do you suggest @Jack-Khuu ?

Jack-Khuu:
Thanks @gabe-l-hart for the PR, taking a deeper look at it now

Seconding what @mikekg mentioned, our workflow currently works as shell scripts (plays well since we run the commands on a suite of various machines), but that mostly helps internal CI. 

I want to add a unittest directory (python backed) that OSS can replicate locally, but haven't got around to it. Might spin up a umbrella laundry list issue so the community can contribute

gabe-l-hart:
Sounds great! I've used `pytest` pretty extensively in the past and would be happy to contribute to tests in the future if needed.

torchchat/cli/convert_hf_checkpoint.py
```diff
@@ -130,31 +136,34 @@ def load_safetensors():
...
             abstract_key = re.sub(r"(\d+)", "{}", key)
             layer_num = re.search(r"\d+", key).group(0)
-            new_key = weight_map[abstract_key]
-            if new_key is None:
-                continue
+            new_key = weight_map.get(abstract_key, abstract_key)
             new_key = new_key.format(layer_num)
         else:
-            new_key = weight_map[key]
+            new_key = weight_map.get(key, key)
```
Jack-Khuu, line 148: I might be misreading something, but why are we keeping keys with missing entries

torchchat/cli/convert_hf_checkpoint.py
```diff
@@ -130,31 +136,34 @@ def load_safetensors():
...
-        if "wq" in key:
+        if "wq.weight" in key or "wq.bias" in key:
+            wk_key = key.replace("wq", "wk")
+            wv_key = key.replace("wq", "wv")
             q = final_result[key]
-            k = final_result[key.replace("wq", "wk")]
-            v = final_result[key.replace("wq", "wv")]
+            k = final_result[wk_key]
+            v = final_result[wv_key]
+            print(key)
```
Jack-Khuu, line 159:
```suggestion
```

torchchat/cli/convert_hf_checkpoint.py
```diff
@@ -130,31 +136,34 @@ def load_safetensors():
...
             q = permute(q, config.n_heads)
+            print(wk_key)
```
Jack-Khuu, line 161:
```suggestion
```

Jack-Khuu (approved):
Nice, thanks for adding

Jack-Khuu:
Hmmm one thing that we'll need to do is tell people to redownload though if they already have a local checkpoint...

we can spin up that message right before we land

torchchat/cli/convert_hf_checkpoint.py
gabe-l-hart: Yikes! Sloppy, good catch

torchchat/cli/convert_hf_checkpoint.py
gabe-l-hart: This is probably not fully needed now that I understand how the mapping is working better. I added this when trying to get it working since Granite Code had missing entries. I was thinking at the time that unconverted tensor names would be treated in a generic way post-conversion, but I now realize that's not the case. I'll remove this change.

torchchat/cli/convert_hf_checkpoint.py
```diff
@@ -130,31 +136,32 @@ def load_safetensors():
...
+
     final_result = {}
     for key, value in merged_result.items():
         if "layers" in key:
             abstract_key = re.sub(r"(\d+)", "{}", key)
             layer_num = re.search(r"\d+", key).group(0)
-            new_key = weight_map[abstract_key]
-            if new_key is None:
-                continue
+            new_key = weight_map.get(abstract_key, abstract_key)
```
Jack-Khuu, line 145: Ditto?

gabe-l-hart: Yep, good catch

Jack-Khuu:
Whoops thought I hit land last night, will land when CI finishes

PR #1260, opened 2024-10-03
Tied word embeddings

gabe-l-hart:
## Dependencies

This PR is part of a sequence in support of adding Granite Code. It depends on merging the following PRs:

- [x] Safetensors: #1255
- [x] Bias tensors: #1259 

## Issues

Closes #1252 

## Description

This PR adds support for models which have shared weights between the input word embeddings and the output layer (tied word embeddings).

## Changes

* Add the `tie_word_embeddings` parameter to `TransformerArgs`
* Add a `load_hook` to the `Transformer` class where, if configured with `tie_word_embeddings`, the `self.output` module will share weights with the `self.tok_embeddings` module.

## Testing

In conjunction with my other changes for Granite Code, I've been able to validate that the results produced with this logic do produce the expected token sequence.

NOTE: If there's any preferred way to include unit tests along with the PR, please let me know and I can get them added! I don't see a familiar unit test structure in the project at this point, so I've been relying on local ad-hoc testing.

PR #1261, opened 2024-10-03
Tokenizers tokenizer

### LINKED ISSUE #1251: Add support for `tokenizers` tokenizers

The request is to extend the [tokenizer](https://github.com/pytorch/torchchat/tree/main/tokenizer) module in `torchchat` to support tokenizers that use the Huggingface [tokenizers](https://github.com/huggingface/tokenizers) library.

There are many models out there that use `tokenizers` which won't be able to run in `torchchat` until they can be loaded and run either via the `tokenizers` library directly or via a conversion to `tiktoken` or `sentencepiece`.

### Alternatives

It may be possible to convert a `tokenizers` tokenizer to a `tiktoken` tokenizer. I have a working implementation of this for the `llama` [tokenizer.json](https://huggingface.co/meta-llama/Llama-3.1-8B-Instruct/blob/main/tokenizer.json) model, however other models that use different `tokenizers` configurations do not work (in particular Granite Code).

### Additional context

This issue is a piece of the puzzle for adding support for Granite Code 3b/8b which use the llama architecture in transormers, but take advantage several pieces of the architecture that are not currently supported by torchchat. The work-in-progress for Granite Code can be found on my fork: https://github.com/gabe-l-hart/torchchat/tree/GraniteCodeSupport.

I have a less fully-fleshed working version of this that I plan to put up as a Draft PR for discussion. I am not intimately familiar with the algorithmic differences between `tiktoken` and the various `tokenizers` pieces (in particular the `pretokenizer`s). My branch has a python implementation that simply wraps `tokenizers`, but I have not yet tried to export Granite Code to other formats where I suspect it would break without a corresponding `c++` implementation. I plan to investigate this further soon!

### RFC (Optional)

_No response_

gabe-l-hart:
## Dependencies

- [x] Safetensors: #1255
- [x] Bias tensors: #1259 
- [x] Tied word embeddings: #1260 

## Issues

Closes #1251 

## Description

This PR adds partial support for models that use the `tokenizers` (as opposed to `tiktoken` or `sentencepiece`) for tokenization. This PR only addresses support in the `python` runner, and it does so by creating a new class in the `tokenizer` module that simply wraps `tokenizers`.

## Discussion

I'm not sure this is the correct direction to go for solving this since the `tokenizers` library is not (to the best of my knowledge) portable to the various export formats (yet). There are two main challenges to extending more tokenizer support outside of simply wrapping `tokenizers`:

### Pre-tokenizers

For may tokenizers, multiple regexes are used in sequence to split the raw string. Not being a regex expert myself, it's not immediately clear to me if it's possible to merge this kind of multi-pass splitting into a single regex. For other tokenizers, a single regex is used, but it is a different expression than any of those [currently implemented in `tiktoken`](https://github.com/pytorch/torchchat/blob/main/tokenizer/tokenizer.h#L138).

From my investigation, I think there are a few candidate paths forward:

1. Provide a `c++` implementation of the various tokenization routines from `tokenizers` in a separate implementation of the [`Tokenizer` class](https://github.com/pytorch/torchchat/blob/main/tokenizer/tokenizer.h#L27).
2. Extend the existing `c++` [`TikToken` class](https://github.com/pytorch/torchchat/blob/main/tokenizer/tiktoken.cpp) to support multiple regexes in the pre-tokenizer
    * This would also correspond with needing to make the set of patterns configurable and therefore serialized into the `tokenizer.model` artifact, or somehow making these tokenizer arguments an argument at instantiation time.

**NOTE:** The corresponding tokenization in `llama.cpp` lives [here](https://github.com/ggerganov/llama.cpp/blob/master/src/llama-vocab.cpp). This code is a full implementation of a unified tokenizer with configuration to dispatch between known patterns and optimized implementations. The config for the model that indicates which tokenizer to use is stored in the model's `GGUF` file directly, so at load time, the correct tokenizer is found based on that value.

### Special Tokens

Even for models that use a single regex (and even the `llama` regex), models may use different special tokens for special functionality (chat template, FIM, tool calling, other custom prompting). Since the `tokenizer.model`, only the vocab is stored, so there is not currently any way to note the special tokens in serialization (similar to the need for configuration of pre-tokenizers).

### DIFF
--- a/torchchat/cli/builder.py
+++ b/torchchat/cli/builder.py
@@ -204,6 +204,7 @@ class TokenizerArgs:
     tokenizer_path: Optional[Union[Path, str]] = None
     is_sentencepiece: bool = False
     is_tiktoken: bool = False
+    is_tokenizers: bool = False
     t: Optional[Any] = None
 
     def __post_init__(self):
@@ -213,6 +214,7 @@ class TokenizerArgs:
             self.t = TiktokenTokenizer(model_path=str(self.tokenizer_path))
             self.is_tiktoken = True
             self.is_sentencepiece = False
+            self.is_tokenizers = False
             return
         except:
             pass
@@ -223,12 +225,25 @@ class TokenizerArgs:
             self.t = SentencePieceProcessor(model_file=str(self.tokenizer_path))
             self.is_tiktoken = False
             self.is_sentencepiece = True
+            self.is_tokenizers = False
+            return
+        except:
+            pass
+
+        try:
+            from tokenizer.tokenizers import TokenizersTokenizer
+
+            self.t = TokenizersTokenizer(str(self.tokenizer_path))
+            self.is_tiktoken = False
+            self.is_sentencepiece = False
+            self.is_tokenizers = True
             return
         except:
             pass
 
         self.is_tiktoken = False
         self.is_sentencepiece = False
+        self.is_tokenizers = False
         self.t = None
         return
 
@@ -240,16 +255,27 @@ class TokenizerArgs:
         if model is None:
             return
 
-        if self.is_tiktoken == self.is_sentencepiece:
+        if len(list(filter(lambda x: x, [self.is_tiktoken, self.is_tokenizers, self.is_sentencepiece]))) != 1:
             raise RuntimeError(f"no tokenizer was found at {self.tokenizer_path}")
 
         is_tiktoken = self.is_tiktoken
         is_sentencepiece = self.is_sentencepiece
+        is_tokenizers = self.is_tokenizers
         use_tiktoken = model.config.use_tiktoken
+        use_tokenizers = model.config.use_tokenizers
+        use_sentencepiece = not (use_tiktoken or use_tokenizers)
 
-        if not (is_tiktoken == use_tiktoken) or not (is_sentencepiece != use_tiktoken):
+        if (
+            (is_tiktoken and not use_tiktoken) or
+            (is_tokenizers and not use_tokenizers) or
+            (is_sentencepiece and not use_sentencepiece)
+        ):
             raise RuntimeError(
-                f"model-specified tokenizer ({tokenizer_setting_to_name(use_tiktoken)}) does not match provided tokenizer ({tokenizer_setting_to_name(is_tiktoken)}) for {model_description}"
+                "model-specified tokenizer ({}) does not match provided tokenizer ({}) for {}".format(
+                    tokenizer_setting_to_name(use_tiktoken, use_tokenizers),
+                    tokenizer_setting_to_name(is_tiktoken, is_tokenizers),
+                    model_description,
+                )
             )
 
         return
@@ -605,5 +631,9 @@ def _initialize_model(
     return model
 
 
-def tokenizer_setting_to_name(tiktoken: bool = False) -> str:
-    return "TikToken" if tiktoken else "SentencePiece"
+def tokenizer_setting_to_name(tiktoken: bool, tokenizers: bool) -> str:
+    if tiktoken:
+        return "TikToken"
+    if tokenizers:
+        return "Tokenizers"
+    return "SentencePiece"
--- a/torchchat/cli/builder.py
+++ b/torchchat/cli/builder.py
@@ -204,7 +204,7 @@ class TokenizerArgs:
     tokenizer_path: Optional[Union[Path, str]] = None
     is_sentencepiece: bool = False
     is_tiktoken: bool = False
-    is_tokenizers: bool = False
+    is_hf_tokenizer: bool = False
     t: Optional[Any] = None
 
     def __post_init__(self):
@@ -214,7 +214,7 @@ class TokenizerArgs:
             self.t = TiktokenTokenizer(model_path=str(self.tokenizer_path))
             self.is_tiktoken = True
             self.is_sentencepiece = False
-            self.is_tokenizers = False
+            self.is_hf_tokenizer = False
             return
         except:
             pass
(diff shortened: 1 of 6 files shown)


gabe-l-hart:
@Jack-Khuu This PR is now the tip of the chain. I've opened it up to review, but I suspect this one will need a lot more discussion than the others. As an FYI, I'm working on a c++ implementation that would support `tokenizers` tokenizers ([branch](https://github.com/gabe-l-hart/torchchat/tree/TokenizersCpp-1251)), but it's slow going with other competing priorities.

gabe-l-hart:
Moving conversation on the various open questions here.

I think I've just discovered part of why converting from `tokenizers` to `tiktoken` format (e.g. with [my script](https://github.com/pytorch/torchchat/issues/1251#issuecomment-2391687601)) is not straightforward.

One of the main differences between the `tokenizer.model` format and the `tokenizer.json`, besides the presence of a bunch of metadata, is that the `vocab` and `merges` are held separately in `tokenizer.json` whereas the merge ranks are explicitly expected to match the IDs in `tokenizer.model`. [This comment](https://github.com/pytorch/torchchat/blob/main/tokenizer/tiktoken.cpp#L158) seems to indicate that this is _one_ way that the vocab can be constructed, but that it is not a _required_ part of the BPE algorithm. This would indicate that `tiktoken -> tokenizers` should work fine, but `tokenizers -> tiktoken` will be much harder because there's no guarantee that this assumption about ranks will be met in an arbitrary `vocab`/`merges` in a `tokenizers` model.

Jack-Khuu:
Pardon the delay: I've been OOO (still am) 
Will take a look when I get back to office

Thanks again!!

gabe-l-hart:
Not a problem at all, I've been distracted on other threads too. I have some partial work towards a native `c++` implementation that supports multiple pre-tokenizer regexes and custom special tokens. At the same time, one of those distracting threads has had me looking more closely at `sentencepiece` and it's _possible_ we could go the route of converting from `tokenizers` -> `sentencepiece` and avoid the need for a full c++ implementation. I'll update as I get more clarity.

torchchat/model.py
```diff
@@ -272,7 +272,9 @@ class TransformerArgs:
     norm_eps: float = 1e-5
     multiple_of: int = 256
     ffn_dim_multiplier: Optional[int] = None
+    # Select the desired tokenizer. Defaults to sentencepiece
     use_tiktoken: bool = False
+    use_tokenizers: bool = False
```
Jack-Khuu, line 277:
```suggestion
    use_hf_tokenizers: bool = False
```

torchchat/model.py
```diff
@@ -329,12 +331,14 @@ class ModelArgs:
     model_type: ModelType
     transformer_args: Dict[str, Dict[str, Any]]
     use_tiktoken: bool
+    use_tokenizers: bool
```
Jack-Khuu, line 334:
```suggestion
    use_hf_tokenizers: bool
```

Jack-Khuu (commented):
Out of scope of this PR (i.e. we'll fix afterwards), but we should probably move toward using an enum for the tokenizer to save us some headache

Jack-Khuu:
Mini Update: We have some engineers internally who may be interested on helping on C++ front if you get stuck btw

gabe-l-hart:
That's great! I'm finally getting back to this. Will push updates for your suggestions and will push a branch with the _very_ WIP c++ stuff.

torchchat/cli/builder.py
```diff
@@ -229,16 +244,27 @@ def validate_model(
         if model is None:
             return
 
-        if self.is_tiktoken == self.is_sentencepiece:
+        if len(list(filter(lambda x: x, [self.is_tiktoken, self.is_tokenizers, self.is_sentencepiece]))) != 1:
```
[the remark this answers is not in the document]
gabe-l-hart: Nice, that's way simpler!

gabe-l-hart (commented):
Just pushed changes with all the renames. Thanks for the suggestion!

gabe-l-hart:
Oops, missed one place to change the name. Should be fixed now.

gabe-l-hart:
_very_ WIP on the `c++` implementation is on a branch now: https://github.com/gabe-l-hart/torchchat/tree/TokenizersCpp-1251

Jack-Khuu (approved):

mikekgfb:
> _very_ WIP on the `c++` implementation is on a branch now: https://github.com/gabe-l-hart/torchchat/tree/TokenizersCpp-1251

Once all the changes needed to support granite have landed, be sure to add the models to the known model json and the readme, please?

Also, at that point when the granite models work with the code that's checked in, is there a smallish granite model (ideally without special license that needs to be accepted, avoiding having to deal with HF tokens as github secrets?) that could be run as end-to-end test?

gabe-l-hart:
>Also, at that point when the granite models work with the code that's checked in, is there a smallish granite model (ideally without special license that needs to be accepted, avoiding having to deal with HF tokens as github secrets?) that could be run as end-to-end test?

All Granite models (starting with the Granite Code ones) are under Apache-2. The smallest Granite Code is [the 3b](https://huggingface.co/ibm-granite/granite-3b-code-base-128k) model which is admittedly not CI/CD sized. Once we start tackling the `"granite"` and `"granitemoe"` architectures (Granite 3.X), the HF team has also created tiny random test models that can be used to ensure the tensors flow, but don't produce any real output.

gabe-l-hart:
For the discussion around `c++` support for HF tokenizers, I recently discovered the tokenizer implementation in [mlc-llm](https://github.com/mlc-ai/mlc-llm/tree/main/cpp/tokenizers) which _may_ be pretty close to exactly what we need. I'm not deeply familiar with the project (thus just finding it), but it could be an interesting starting point for a general-purpose `c++` tokenizer solution. Another option would be to look into hoisting the [`llama.cpp` tokenization](https://github.com/ggerganov/llama.cpp/blob/master/src/llama-vocab.h) into its own project, but that would likely require some significant untangling from their codebase and would be hard to maintain as the project evolves.

PR #1266, opened 2024-10-04
[Distributed] add TokenizerType enum, spelling

lessw2020:
This PR is minor:
1 - adds an enum for TokenType so that we can better handle various tokenizer types directly. 
2 - renames decorder to decoder for clarity.

### DIFF
--- a/dist_run.py
+++ b/dist_run.py
@@ -10,6 +10,7 @@
 
 import argparse
 import os
+from enum import auto, Enum
 from pathlib import Path
 from types import SimpleNamespace
 from typing import Any, Dict, List, Optional, Tuple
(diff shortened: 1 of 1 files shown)


dist_run.py
```diff
@@ -294,7 +309,7 @@ def main(args):
     config = TransformerArgs.from_params(model_config.transformer_args["text"])
     logger.info(f"Transformer Config: {config}")
 
-    tokenizer = _build_chat_tokenizer(model_name)
+    tokenizer, tokenizer_type = _build_chat_tokenizer(model_name)
```
kwen2501, line 312: nit: can you make `tokenizer_type` a global so that we can also use it in functions like `_decode_in_flight` ?
`_build_chat_tokenizer` will set this global, and will not need to return it.

kwen2501 (approved):
Thanks. LGTM. Added a comment.

dist_run.py
lessw2020: I've updated to make tokenizer_type a global (and renamed to _tokenizer_type as a type hint to it's global state).

PR #1268, opened 2024-10-04
[Distributed] Add support for torchchat checkpoint format

kwen2501:
Distributed workflow supports two formats now, controlled by the `--chpt-from` flag ("hf" or "torchchat").
The difference is whether there is an index file -- impacts loading in pipeline parallel case.

```
    if chpt_from == "hf":
        # This format stands for: index file + multiple binary files

    elif chpt_from == "torchchat":
        # This format stands for:
        # single binary file, OR
        # multiple binary files without index files.
```

`torchrun --standalone --nproc-per-node 4 dist_run.py llama3 --pp 2 --chpt-from hf`
`torchrun --standalone --nproc-per-node 4 dist_run.py llama3 --pp 2 --chpt-from torchchat`
both run.
But they generate different responses. To be investigated further.

TODO:
we should create real model instead of creating Transformer.


mikekg:
Can you figure out from what's in the location whether it's one or the other format rather than adding another option? (If push comes to shove, you could expect the checkpoint path point to the index file?)

Also, does this compose w/ #1255?

kwen2501:
@mikekg Good idea. Should be easy to implement.

dist_run.py
```diff
@@ -110,26 +110,33 @@ def _build_chat_tokenizer(
     return tokenizer
 
 
-def _load_model_weights(stage_module, distribution, device, model_config):
+def _load_model_weights(
+    stage_module: torch.nn.Module,
+    distribution: str,
+    device: torch.device,
+    model_config: ModelArgs,
+    chpt_from: str,
```
lessw2020, line 118: for better clarity, 'chkpt' is a much better abbreviation for checkpoint.  chpt is short for 'chapter' which is confusing.

torchchat/distributed/checkpoint_utils.py
```diff
@@ -264,64 +266,66 @@ def permute_weight_to_attn_heads(w, n_heads, head_dim, model_dim):
...
+    updated_states: Optional[Set[str]]= None,
 ):
-    count_dtensors_loaded = 0
+    """
+    Update the state dict with the checkpoint tensors.
+    Note:
+    - For HF format, `new_to_old_keymap` is a mapping from the new key to the old
+    key.
+    - For torchchat format, `new_to_old_keymap` is None (because FQN conversion
+    has been doen by torchchat download script).
```
lessw2020, line 280: doen = done :)

torchchat/distributed/checkpoint_utils.py
```diff
@@ -366,3 +370,83 @@ def log_loading_status(missing_keys: Set[str], updated_states: Set[str]):
...
+        "checkpoint_dir": checkpoint_dir,
+        "checkpoint_path": checkpoint_path,
+    }
+    builder_args = BuilderArgs(**args_dict)
+    # Then, load the checkpoint using torchchat util
+    checkpoint = _load_checkpoint(builder_args)
+
+    updated_states: Set[str] = set()
+    # This step converts full tensor into DTensor
+    update_state_dict(
```
lessw2020, line 442: need to pass in if this is torchchat or hf to permute or not?

torchchat/distributed/checkpoint_utils.py
```diff
@@ -264,64 +266,66 @@ def permute_weight_to_attn_heads(w, n_heads, head_dim, model_dim):
...
+            old_param = param
+
+        if old_param not in checkpoint:
+            # Maybe this param is in other files
+            continue
+
+        checkpoint_tensor = checkpoint[old_param]
+        model_tensor = state_dict[param]
+
+        if "wq" in param:
```
lessw2020, line 307: need to check what you are loading here.  
torchchat won't need permuting.

lessw2020 (approved):
approving to land. 
I think your code here doesn't work b/c you are not checking what type of state dict you are updating.  Thus it seems you are going to permute a torchchat checkpoint which doesn't need it. 
That may be why: "But they generate different responses. To be investigated further."

Also, please switch the abbreviation for checkpoint to "chkpt' and not "chpt".  'Chpt' to me = 'chapter' and reads confusingly :)

PR #1271, opened 2024-10-04
[Distributed] Add a shard helper for creating DTensors

kwen2501:
Adding the same `shard` function as used in torchao's TP example:
https://github.com/pytorch/ao/pull/937

Tried upstream the helper to pytorch, but the [PR](https://github.com/pytorch/pytorch/pull/136589) is going nowhere. 
So hosting here for now.


torchchat/distributed/dtensor_utils.py
```diff
@@ -1,77 +1,69 @@
...
-        if placement.type == "shard":
-            logger.info(f"  Sharding spec: {placement.sharding_spec}")
-        elif placement.type == "replicate":
-            logger.info("  Replicated across devices")
-        else:
-            logger.info(f"  Other placement type: {placement.type}")
-
-    logger.info(f"Device mesh shape: {dtensor.device_mesh.shape}")
-    logger.info(f"Device mesh devices: {dtensor.device_mesh.device_type}")
+def shard(
```
lessw2020, line 35: minor nit - having a function called shard when we have the imported Shard is asking for confusion (same name, one letter cap difference). 
better name here might be shard_full_tensor?

lessw2020 (approved):
lgtm! 
minor nit on potential naming improvement for shard.

PR #1275, opened 2024-10-05
[Distributed] Support index + multi-bin loading

kwen2501:
Some HF format uses index + multiple binaries instead of index + multiple safetensors.
Distributed workflows already support the later, this PR adds support for the former too. 
(The reason is that quantized checkpoint cannot save as safetensor format yet.)

### DIFF
--- a/torchchat/distributed/checkpoint_utils.py
+++ b/torchchat/distributed/checkpoint_utils.py
@@ -19,7 +19,8 @@ from torchchat.distributed.dtensor_utils import convert_to_dtensor
 from torchchat.cli.builder import BuilderArgs, _load_checkpoint
 
 
-_DEFAULT_SAFETENSOR_FILE_NAME = "model.safetensors.index.json"
+_DEFAULT_SAFETENSOR_INDEX = "model.safetensors.index.json"
+_DEFAULT_BIN_INDEX = "pytorch_model.bin.index.json"
 _CONFIG_NAME = "config.json"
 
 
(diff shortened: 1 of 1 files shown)


mikekg:
Can you figure out from what's in the location whether it's one or the other format rather than adding another option?  (If push comes to shove, you could expect the checkpoint path point to the index file?)

mikekg:
Does this compose with #1255  ?

kwen2501:
> Does this compose with #1255 ?

Thanks for the review. I hope yes. 
Distributed already have support for "index + multi-safetensor" format (thanks to @lessw2020 ). 
This PR only **adds** support for "index + multi-bin" format. (The reason is that quantized checkpoint cannot save as safetensor format yet.)

mikekg:
> > Does this compose with #1255 ?
> 
> Thanks for the review. I hope yes. Distributed already have support for "index + multi-safetensor" format (thanks to @lessw2020 ). This PR only **adds** support for "index + multi-bin" format. (The reason is that quantized checkpoint cannot save as safetensor format yet.)

How much of this can be shared with local-only weight loading?  Should distributed weight loading just be a general case of local loading (or vice versa, local just a special case of distributed?)  Not suggesting to abandon this PR at all, but similar to what you've done to merging model.py and model_dist.py (yay! Awesome work!!!!) could be a real value here.

lessw2020 (approved):
lgtm!
Thanks for adding this!

PR #1276, opened 2024-10-06
[Distributed] Fix correctness issue in TC load path

kwen2501:
torchchat checkpoints are already permuted before saving, thus we don't need to permute it again when loading them.

Also added support for:
`--chpt-from tc:<model_dir>`
Example:
`tc:meta-llama/Meta-Llama-3-8B-Instruct-int8_wo`
(`int8_wo` stands for a quantized checkpoint)


torchchat/distributed/checkpoint_utils.py
```diff
@@ -118,14 +118,12 @@ def remap_weight_keys(dictionary):
 
 
 def load_weights_per_map(
-    stage_module: Module,
+    stage_state_dict,
```
lessw2020, line 121: typing is missing?

torchchat/distributed/checkpoint_utils.py
```diff
@@ -287,14 +260,15 @@ def update_state_dict(
...
-        if "wq" in param:
-            checkpoint_tensor = permute_weight_to_attn_heads(
-                checkpoint_tensor, num_heads, head_dim, dim
-            )
-        elif "wk" in param:
-            checkpoint_tensor = permute_weight_to_attn_heads(
-                checkpoint_tensor, num_local_heads, head_dim, dim
-            )
+        if new_to_old_keymap is not None:
+            if "wq" in param:
```
lessw2020, line 264: so this was the issue!  It seems like a flag as to the type of checkpoint is more robust and then map type to need for permuting?  Seems brittle if we have future checkpoints and we permute or not solely based on a fqn (i.e. generically, having 'wq' in a name doesn't actually imply anything inherently about need to permute)?

dist_run.py
```diff
@@ -125,18 +125,39 @@ def _load_model_weights(
         distribution (str): The distribution name, e.g. "meta-llama/Meta-Llama-3-8B-Instruct".
         device (torch.device): The device to load the weights onto.
         model_config (ModelArgs): The model config.
-        chpt_from (str): The checkpoint format to load the weights from, e.g. "torchchat" or "hf".
+        chpt_from (str): The checkpoint format to load the weights from.
```
lessw2020, line 128: as in prev comments, pls switch to 'chkpt' for abbreviating checkpoint.  It reads as 'chapter_from' when using 'chpt_from'.

lessw2020 (approved):
thanks for fixing the permute issue.
looks good - left a comment about robustness of relying solely on 'wq' in a param name as the determining factor to permute or not...seems better to map it to checkpoint type directly, for future robustness.

torchchat/distributed/checkpoint_utils.py
kwen2501: I am using `if new_to_old_keymap is not None` as the flag for now.
If we are using original HF checkpoints, then we will need this `new_to_old_keymap`. Otherwise, if we use TC checkpoints, we don't.

PR #1280, opened 2024-10-08
[Not for land] Util for saving quantized model

kwen2501:
When composing distributed with quantization, one potential case is that the model has been quantized and saved so a second run do not need to quantize it again.

This is a script for quantizing a model and save it to disk.

The quantization can be done via either:
- `--quantize abc.json`, or
- `quantize_(model, int8_weight_only())`.

The latter is not in torchchat today.

Cc: @Jack-Khuu @jerryzh168

### DIFF
--- a/torchchat/save_quant.py
+++ b/torchchat/save_quant.py
@@ -1,0 +1,58 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+
+import os
+from pathlib import Path
+from typing import Optional
+
+import torch
+import torch.nn as nn
+
+from torchchat.cli.builder import (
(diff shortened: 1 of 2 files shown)


torchchat/save_quant.py
```diff
@@ -0,0 +1,58 @@
...
+import torch.nn as nn
+
+from torchchat.cli.builder import (
+    _initialize_model,
+    BuilderArgs,
+)
+
+from torchchat.utils.build_utils import set_precision
+
+from torchao.quantization import quantize_, int8_weight_only
```
jerryzh168, line 21: we now support tensor parallelism for int4_weight_only, float8_weight_only and float8_dynamic_activation_float8_weight as well I think, feel free to try out

jerryzh168 (approved):
LGTM

Jack-Khuu:
Closing for now, may revive as reference

Issue #1277, opened 2024-10-06
Android demo app poor model performance


fran-aubry:
### 🐛 Describe the bug

I wanted to try the new Llama 3.2 1B parameter model on mobile. I downloaded the model and generated the `pte` like so:

```
python torchchat.py download llama3.2-1b
python torchchat.py export llama3.2-1b --quantize torchchat/quant_config/mobile.json --output-pte-path llama3_2-1b.pte
```

Then I pushed `llama3_2-1b.pte` file and `tokenizer.model` files to the mobile phone using `adb`. 

I executed the demo app in `torchchat/edge/android/torchchat` using Android Studio with `.aar` file provided on the TorchChat repo readme.

However, when I chat with the AI its responses are very useless and feel quite different than what I get with the same prompt on my computer:

![example](https://github.com/user-attachments/assets/8e9d7128-6afd-46b5-8c1f-6b03ad3bccbb)
![terminal-interaction](https://github.com/user-attachments/assets/beb9733a-3b23-40e7-9354-43a97cb05fa0)

Is there a problem with the default quantization parameters? I tried to not quantize but then the app crashed when loading the model.

### Versions

Collecting environment information...
PyTorch version: 2.5.0.dev20240901
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 14.4 (arm64)
GCC version: Could not collect
Clang version: 15.0.0 (clang-1500.3.9.4)
CMake version: version 3.30.4
Libc version: N/A

Python version: 3.10.0 (default, Mar  3 2022, 03:54:28) [Clang 12.0.0 ] (64-bit runtime)
Python platform: macOS-14.4-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M2 Pro

Versions of relevant libraries:
[pip3] executorch==0.5.0a0+286799c
[pip3] numpy==1.26.4
[pip3] torch==2.5.0.dev20240901
[pip3] torchao==0.5.0+git0916b5b
[pip3] torchaudio==2.5.0.dev20240901
[pip3] torchsr==1.0.4
[pip3] torchtune==0.3.0.dev20240928+cpu
[...]

PR #1284, opened 2024-10-09
Add Prompt Template for Android Demo App

Jack-Khuu:
Adds a basic template to the Android Demo app to simulate a chat response.

Without the template, llama models default to a basic completion inference which isn't the behaviors users expect when using the demo app

---
I would love for someone to verify on device that they are seeing expected behavior (and sharing a screenshot)


Jack-Khuu:
https://github.com/pytorch/torchchat/issues/1277

PR #1291, opened 2024-10-10
Implement the AO API in torchchat quantization handlers and unify logic.

mikekgfb:
Implement the AO API in torchchat quantization handlers and unify logic.

1 - implement .quantize() for TC quantization handlers and support args to make consistent with AO
2 - remove special handling for various combinations of parameters and use validate_args before calling with **q_kwargs
3 - remove check probing whether we successfully loaded a8wx and install an error-reporting handler if loading failed which will be called as quant handler and issue an error
4 - unify both tc and ao quantization handler dicts with shared calling logic
5 - provide informational message when quantizer option not supported (via introspection)


torchchat/utils/quantize.py
```diff
@@ -890,7 +901,7 @@ def quantized_model(self) -> nn.Module:
     sys.modules["torchao_experimental_quant_api"] = torchao_experimental_quant_api
     torchao_experimental_quant_api_spec.loader.exec_module(torchao_experimental_quant_api)
     from torchao_experimental_quant_api import Int8DynActIntxWeightQuantizer
-    ao_quantizer_class_dict["linear:a8wxdq"] = Int8DynActIntxWeightQuantizer
+    quantizer_class_dict["linear:a8wxdq"] = Int8DynActIntxWeightQuantizer
```
larryliu0820, line 904: Any reason we can't put this inside of the `quantizer_class_dict`?

larryliu0820: It seems cleaner to try import `Int8DynActIntxWeightQuantizer` first and fallback to `ErrorHandler`, then assign this handler to `linear:a8wxdq` in `quantizer_class_dict`.

larryliu0820 (changes requested):
Thanks for the cleanup! Just one comment and please make sure CI jobs are passing

mikekgfb:
> Thanks for the cleanup! Just one comment and please make sure CI jobs are passing

mikekgfb:
> > Thanks for the cleanup! Just one comment and please make sure CI jobs are passing

My pleasure.  Sorry about that error, should be fixed now.

torchchat/utils/quantize.py
larryliu0820: I think if this can go away soon I don't have a strong opinion and can live with this.

larryliu0820 (approved):
CI is passing

PR #1299, opened 2024-10-15
Fix dangling open skip in README.md

mikekgfb:
1 - Fix an extraneous skip end that is out of order with a skip begin. 
2 - fix some typos

PS: This might cause some README tests to fail, as they have not been run in a long time (due to misplaced skip causing some commands from the README.md to be skipped during the test).


vmpuri (approved):
Thanks for these changes. We should look into adding a spellcheck linter for comments and docs.

PR #1305, opened 2024-10-16
Remove apparently-unused import torchvision in model.py

swolchok:


vmpuri (approved):
lgtm

mikekgfb:
Should this also include removing torchvision from `install/install_requirements.sh` ?

PR #1307, opened 2024-10-16
Unbreak gguf util CI job by fixing numpy version

larryliu0820:
Setting numpy version to be the range required by gguf: https://github.com/ggerganov/llama.cpp/blob/master/gguf-py/pyproject.toml


swolchok (approved):
yep looks like we have to do this :\

PR #1308, opened 2024-10-16
Minor cleanup in torchchat/cli/builder.py

swolchok:
Beautify a series of similar checks & fix a spelling error.


Jack-Khuu (approved):
Thanks for the cleanup

cc: @kwen2501 on parallelize_model type

PR #1310, opened 2024-10-17
Update install_requirements.sh

mikekgfb:
After #1305 removed dependence on torchvision, this PR drops torchvision installation as part of install_requirements.sh


Jack-Khuu:
Looks like torchtune is looking for torchvision, but isn't being pulled it in as part of tune itself: https://github.com/pytorch/torchchat/actions/runs/11409026850/job/31753111489?pr=1310

@ebsmothers Is this intended behavior or have we imported tune improperly?

mikekgfb:
Never mind, we need torchvision for torchtune (and importing it transitively), so removing the only import of torchvision in model.py is not sufficient to be able to remove torchvision from the list of packages we depend on.

PR #1312, opened 2024-10-17
Install ET nightly and bump up ET version to 20241101

larryliu0820:
Summary:

Change the way we install ExecuTorch python library. We read `et-pin.txt` and get back a string looks like `2024-11-01`. Then we process it and download nightly version `0.5.0.dev20241101`. We also infer the commit hash from this nightly version and use it to pull ET sources.

Test Plan: Rely on CI jobs.

Reviewers:

Subscribers:

Tasks:

Tags:


larryliu0820:
Waiting for a new ET nightly

PR #1317, opened 2024-10-18
Update Android screenshot

kirklandsign:


vmpuri (approved):
Just a nit: are we able to change the filename? `chat.jpg` is quite generic - if the old app image is outdated anyways, are we able to retain the same filename and just delete the old image?

kirklandsign:
> Just a nit: are we able to change the filename? `chat.jpg` is quite generic - if the old app image is outdated anyways, are we able to retain the same filename and just delete the old image?

That's because we have already used chat.png in ET wiki 🤣

Maybe we update, both ET and torchchat, next time when we revisit this?

Jack-Khuu (approved):

PR #1329, opened 2024-10-25
[WIP] Initial PR for generating and loading state dict

Jack-Khuu:


torchchat/cli/cli.py
```diff
@@ -148,6 +148,12 @@ def _add_model_config_args(parser, verb: str) -> None:
             help="Whether to compile the prefill. Improves prefill perf, but has higher compile times.",
         )
 
+    model_config_parser.add_argument(
```
jerryzh168, line 151: after migration we shouldn't need anything special for quantized model right

Jack-Khuu: Not sure I follow? The things I'm testing out should work for tensor subclass right?

jerryzh168: oh I meant that with tensor subclass API, quantized checkpoint should be able to be loaded the same way as normal checkpoint.

the code path of loading a quantized model v.s. quantizing model on the fly might still make sense though, maybe just need to change the naming or something

Jack-Khuu:
Closing for now, may revive as reference

PR #1335, opened 2024-10-30
[distributed] Add Llama3-70B for distributed inference

lessw2020:
this PR:
1 - adds llama3-70B as another supported model for dist inference.  
2 - updates the model_dims to auto adjust based on model in our master dict (i.e. 4096 vs 8192). 

Screenshot 2024-10-29 at 5 13 19 PM


mreso (approved):
LGTM

Issue #1041, opened 2024-08-19
Improve support for and documentation of custom models


Jack-Khuu:
### 🚀 The feature, motivation and pitch

torchchat supports adding models to the "known_model" list and has CLI support for local models not hosted in torchchat's, but this can be better documented. 

### Additional context

Some PR's Related to this theme:
* https://github.com/pytorch/torchchat/issues/1038 
* https://github.com/pytorch/torchchat/issues/1040

Jack-Khuu:
https://github.com/pytorch/torchchat/blob/main/docs/local-model.md

More work to come this half, closing issue

Issue #1038, opened 2024-08-16
How to deploy a new model by torchchat?


liu8060:
I want to use torchchat to load the trained model directly from the local. How to change the torchchat/config/data/models.json? Need to change download _ and _ convert in download.py?And, what other documents may need to be changed?

Jack-Khuu:
Glad to have you try things out

What file format is the local model you're working with?

liu8060:
two formats
One is a folder format, which includes the model’s weight files, tokenizer files, and other necessary components. 
The other format is a PyTorch .pth file, which contains the entire model.

The model's base model is llama3-8b

Jack-Khuu:
If the model is accessible from huggingface: Here's an example PR of how you can add it

https://github.com/pytorch/torchchat/pull/947 Specifically the known_model_configs and model.json


> The other format is a PyTorch .pth file, which contains the entire model.

For this you can can add a known_model_config (based on your params.json with "use_tiktoken"=true for llama3 derivatives) and then point to your pth with `--checkpoint-path`

For example: `python torchchat.py generate --checkpoint-path ~/Desktop/test/consolidated.00.pth --prompt "What should I do in Manhattan Chinatown?" --device mps --num-samples 3`

Jack-Khuu:
> One is a folder format, which includes the model’s weight files, tokenizer files, and other necessary components.

Similar to this case? https://github.com/pytorch/torchchat/issues/1040

PR #1336, opened 2024-10-31
Granite code support

### LINKED ISSUE #1262: Support Granite Code 3B/8B

The `torchchat` framework provides an excellent platform for embedding models into many different edge-centric platforms.

The [Granite Code models](https://huggingface.co/collections/ibm-granite/granite-code-models-6624c5cec322e4c148c8b330), specifically the [3B-128k](https://huggingface.co/ibm-granite/granite-3b-code-instruct-128k) and [8B-128k](https://huggingface.co/ibm-granite/granite-8b-code-instruct-128k) variants, are a family of models from IBM that support a wide variety of code-related tasks. The models are released under the Apache-3 license and are therefore well-suited to embedded use-cases where code intelligence is needed. 

The request here is to extend the model support in `torchchat` to support running the 3B and 8B long-context variants of Granite Code in order to enable usage of these models across embedded use-cases.

### Alternatives

Depending on the goals of the `torchchat` framework, extending support to non-llama models may or may not be a project goal. There are other embedded frameworks out there (notably `llama.cpp` and the many projects that wrap it), so these can be used to run Granite Code in embedded environments. Our goal at IBM is to provide users with as many choices as possible on how to run all of our Granite family models, so our hope is that `torchchat` can be a strong piece of this story!

### Additional context

The 3B and 8B models use the `llama` architecture in `transformers`, so they are _close_ to fully supported as-is. There are a few crucial pieces that are present in the `transformers` implementation that are missing in `torchchat`:

* Safetensors support: https://github.com/pytorch/torchchat/issues/1249
* Tied word embeddings: https://github.com/pytorch/torchchat/issues/1252
* Bias tensors: https://github.com/pytorch/torchchat/issues/1250
* Non-tiktoken/sentencepiece tokenizers: https://github.com/pytorch/torchchat/issues/1251

### RFC (Optional)

I've worked through the initial steps of solving all of these outstanding issues (see the corresponding issues). Once these are solved, the addition of these Granite Code models should consist of the following steps:

* Adding new entries to [models.json](https://github.com/pytorch/torchchat/blob/main/torchchat/model_config/models.json)
* Adding the right set of model-specific params to [model_params](https://github.com/pytorch/torchchat/tree/main/torchchat/model_params)

gabe-l-hart:
## Dependencies

- [x] Safetensors: #1255
- [x] Bias tensors: #1259 
- [x] Tied word embeddings: #1260 
- [x] Tokenizers tokenizer: #1261

## Issues

Closes #1262 

## Description

This PR adds support for Granite Code in 3B and 8B sizes. Given current limitations with the export of tokenizers, they will only work in the python environment with this PR.

### Usage

To test using these models, I did it both by running with the aliases and by running pointing directly at the checkpoint/tokenizer:

```sh
# Run with alias
python torchchat.py generate granite-code \
  --prompt "Write a python function to sort numbers and strings with numeric prefixes"

# Run with direct reference to artifacts
python torchchat.py generate \
  --prompt "Write a python function to sort numbers and strings with numeric prefixes" \
  --checkpoint-path $HOME/models/ibm-granite/granite-3b-code-instruct-128k/model.pth \
  --tokenizer-path $HOME/models/ibm-granite/granite-3b-code-instruct-128k/tokenizer.json \
  --params-path torchchat/model_params/Granite-3B-Code.json
```

### Open Questions

There are several outstanding issues, beyond the upstream tokenizers PR, that need to be solved before this PR is ready for full review:

- [x] It seems that in `chat` mode, the models produce very unreliable results, sometimes generating a single token while other times generating a reasonable result but stopping mid-sentence before reaching the max token limit. My current hypothesis is that the chat template is not being used anywhere and we're therefore using the `llama` chat template automatically.
- [ ] The 8B model currently produces garbage after a few tokens. The main difference between the 3B and 8B models, besides common parameter differences like number of layers and hidden size, is that the 8B uses grouped query attention. I've seen similar behavior in other frameworks where the model starts on a good track, then devolves into garbage and in those cases GQA was also at play, so I suspect it's something along these lines here as well.


gabe-l-hart:
Also, I used the following script to perform conversion of a pre-existing HF snapshot. It's similar to the `if __name__ == "__main__"` block in `convert_hf_checkpoint.py`:

<details>
<summary>convert_existing_checkpoint.py</summary>

```py
#!/usr/bin/env python
"""
Simple script to convert an existing HF snapshot into torchchat format
"""

# Standard
import argparse
from pathlib import Path

# Local
from torchchat.cli.convert_hf_checkpoint import convert_hf_checkpoint, convert_hf_checkpoint_to_tune

def main():
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument("checkpoint_dir", help="Directory containing HF checkpoint")
    parser.add_argument("--name", "-n", default=None, help="Name to use for the model")
    parser.add_argument("--torchtune", "-t", action="store_true", default=False, help="Convert to torchtune format")
    args = parser.parse_args()
    if args.torchtune:
        convert_hf_checkpoint_to_tune(model_dir=Path(args.checkpoint_dir), model_name=args.name)
    else:
        convert_hf_checkpoint(model_dir=Path(args.checkpoint_dir), model_name=args.name)

if __name__ == "__main__":
    main()
```

</details>

gabe-l-hart:
I confirmed that it was falling back to the `llama2` chat formatter because it wasn't using `tiktoken`. I've added basic `jinja2` chat template support when using the HF tokenizer.

mikekgfb:
A pointer to this PR and the example commands from the PR description would make a good starting point for docs/new_model.md to (at least partially?) address #1038 / #1041 in conjunction with some explanatory text

```
# wget artifacts here
# Run with direct reference to artifacts
python torchchat.py generate \
  --prompt "Write a python function to sort numbers and strings with numeric prefixes" \
  --checkpoint-path $HOME/models/ibm-granite/granite-3b-code-instruct-128k/model.pth \
  --tokenizer-path $HOME/models/ibm-granite/granite-3b-code-instruct-128k/tokenizer.json \
  --params-path torchchat/model_params/Granite-3B-Code.json
```

Explain how to add to model list....

```
# Run with alias
python torchchat.py generate granite-code \
  --prompt "Write a python function to sort numbers and strings with numeric prefixes"
```

if added to `.ci/scripts/run-docs new_model` it might also make a testcase for the features used in granite.  I would suggest to use the 1B model for faster test time, run-docs can substitute 3b -> 1b model, similar to how llama in the docs is replaced with stories for tests.

gabe-l-hart:
@Jack-Khuu I'm a bit stumped trying to get the 8B model working. I'm trying to mentally diff the Attention implementation in [torchchat](https://github.com/pytorch/torchchat/blob/main/torchchat/model.py#L845) vs [transformers](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L345) to see if I can find anything that would indicate something behaving differently with Grouped Query Attention.

I'm not really following the different way that the `torchchat` version is manipulating the tensors for tensor parallel inference (need to do some background reading there), but this _feels_ like it's got to be close to the root of the issue. The only other place that I could imagine things going wrong is in the unpacking of the unified `wqkv` [here](https://github.com/pytorch/torchchat/blob/main/torchchat/model.py#L788). Any insight you can offer would be much appreciated!

<details>
<summary>Results with 3B</summary>


```sh
?> python torchchat.py generate granite-code-3b --prompt "Write a python hello world function"
PyTorch version 2.6.0.dev20241002 available.
W1108 13:18:36.747000 52813 site-packages/torch/distributed/elastic/multiprocessing/redirects.py:29] NOTE: Redirects are currently not supported in Windows or MacOs.
Using device=mps 
Loading model...
Time to load model: 3.86 seconds
Write a python hello world function

`​``python
def say_hello():
    print("hello world")
`​``

Generated 19 tokens                 
Time for inference 1: 1.6639 sec total                 
Time to first token: 0.4289 sec with parallel prefill.                

      Total throughput: 12.0199 tokens/sec, 0.0832 s/token                 
First token throughput: 2.3316 tokens/sec, 0.4289 s/token                 
 Next token throughput: 15.3844 tokens/sec, 0.0650 s/token                     

Bandwidth achieved: 86.74 GB/s

      Average tokens/sec (total): 12.02                 
Average tokens/sec (first token): 2.33                 
Average tokens/sec (next tokens): 15.38 
```

</details>

<details>
<summary>Results with 8B</summary>


```sh
?> python torchchat.py generate granite-code-3b --prompt "Write a python hello world function"
PyTorch version 2.6.0.dev20241002 available.
W1108 13:18:36.747000 52813 site-packages/torch/distributed/elastic/multiprocessing/redirects.py:29] NOTE: Redirects are currently not supported in Windows or MacOs.
Using device=mps 
Loading model...
Time to load model: 3.86 seconds
Write a python hello world function

`​``python
def say_hello():
    print("hello world")
`​``

Generated 19 tokens                 
Time for inference 1: 1.6639 sec total                 
Time to first token: 0.4289 sec with parallel prefill.                

      Total throughput: 12.0199 tokens/sec, 0.0832 s/token                 
First token throughput: 2.3316 tokens/sec, 0.4289 s/token                 
 Next token throughput: 15.3844 tokens/sec, 0.0650 s/token                     

Bandwidth achieved: 86.74 GB/s

      Average tokens/sec (total): 12.02                 
Average tokens/sec (first token): 2.33                 
Average tokens/sec (next tokens): 15.38 
```

</details>

Jack-Khuu:
Thanks for the details @gabe-l-hart 

I'll try to give it a gander this weekend. It's weird that 3B works, but 8B doesn't. I assume they use the same template so that at least clears that part

byjlw:
Looks like this has been open for a several weeks now. 
Yeah, the template thing is super hacky right now and I knew it was going to hang up our ability to add new models. 
In general we need to make a smoother path for new adding new models with different architectures, templates and storage locations. 

It's been on @varunfb  and @Jack-Khuu 's plate for a while but they've been swamped with other work. 
Fortunately it's planning season and the design for this is on the list. 
@gabe-l-hart would love to get your feedback on how best to support folks like yourself.

gabe-l-hart:
Thanks @byjlw! I definitely understand juggling priorities. The path to adding new models in the `model_params` and `model_config` is relatively straightforward (could use a doc, but TBH I never read docs anyway, so easy-to-read code is always best). The real challenge has come up around places where the models differ from the `llama` series models. In particular, Granite Code uses the `llama` architecture, but uses several optional bits that the Meta Llama models don't (e.g. HF tokenizers, tied embeddings). Getting these pieces to work has been a decently steep learning curve (fun though!). I think the thing that would be most helpful would be some kind of compatibility matrix doc that shows architectures that have support, sub-features within architectures, and which "layers" they're supported in (e.g. python, c++, executorch). This would help a lot in figuring out where to dig in to add new model support.

For the specific issues for Granite Code, the place I'm a bit stuck is trying to figure out why the `8B` model is flopping while the `3B` model is working just fine. My gut is that it has something to do with the alternate attention mechanism in TC, but I'm not deeply versed in attention enough to spot it quickly. The only architectural difference between 3B and 8B is the use of grouped query attention, so it's either something there or there's some incompatibility between the attention implementations in `transformers` and TC that's only being exercised by the specific weights of the 8B. Any help and/or expert bug spotting would be much appreciated!

gabe-l-hart:
I just rebased on `main` and it now looks like even the 3b model is producing only a single token as output. Will try to get to the bottom of it.

mikekgfb:
> I just rebased on `main` and it now looks like even the 3b model is producing only a single token as output in `chat` mode. Will try to get to the bottom of it.

Have you tried bisecting the 3B fail?  Even if the change was legit and necessary, the type of change that would break the 3B model might give insight in how to "fix" both the 3B and 8B models?
.

mikekgfb:
> The real challenge has come up around places where the models differ from the `llama` series models. In particular, Granite Code uses the `llama` architecture, but uses several optional bits that the Meta Llama models don't (e.g. HF tokenizers, tied embeddings). Getting these pieces to work has been a decently steep learning curve (fun though!).

I'm a bit surprised by this because chatgpt had this to say (understanding that I'm quoting chatgppt about an IBM model to an IBMer, so skating on seriously thin ice!!!):

what tokenization scheme does the ibm granite model use

Searched 4 sites
The IBM Granite models, including its base and instruction-tuned variants, utilize the Llama2 tokenizer for tokenization. This choice aligns with the models’ architectural similarity to Meta's Llama2 series, such as the Granite-7b model, which follows the Llama2-7B architecture and employs similar tokenization strategies. These tokenizers are designed to handle diverse data sources, including programming languages and natural language, ensuring compatibility and efficiency in tasks like code synthesis and language understanding​

So in theory, SentencePiece should do the trick?  Is it the pre and post processing with regexps? (I think I saw some discussion about regexps in one of your PRs or issues?)

gabe-l-hart:
>Have you tried bisecting the 3B fail? Even if the change was legit and necessary, the type of change that would break the 3B model might give insight in how to "fix" both the 3B and 8B models?

That's on my todo list for my next chunk of uninterrupted dev time! I'm hoping that will be today.

>I'm a bit surprised by this because chatgpt had this to say (understanding that I'm quoting chatgppt about an IBM model to an IBMer, so skating on seriously thin ice!!!):

Heh, as you know I'm sure, IBM is a _big_ place, so I'm definitely doing a lot of learning myself in this space. My info from the models team is that we've been using the `starcoder` tokenizer up until now (including `Granite Code` and the `Granite 3.0` series). When first trying to understand how best to support that in `torchchat`, I was missing a lot of knowledge about `sentencepiece`, so was working off of the `tokenizer_config.json` in HF. I suspect it _would_ be possible to reverse-convert from `tokenizers` back to `sentencepiece` for this config, but I haven't done that work yet since I was already halfway down the rabbit hole of `tokenizers` support. We can certainly look into that as an alternative approach if the preference is to avoid the complexity of the c++ tokenizer buildout.

gabe-l-hart:
@Jack-Khuu @mikekgfb @byjlw I figured out where the issues were coming from. It was two things:

1. The logic was always inserting a `bos` token at the beginning of the sequence which the 3b model was _sometimes_ ok with, but the 8b never was
    * To solve this, I added `tokenizer_prepend_bos` as a parameter in `TransformerArgs` and `ModelArgs`. It seemed a little klunky to plumb it through multiple blobs, but this got things working for both models with raw generation
2. The chat template logic was not robust beyond `llama2` and `llama3` templating. Solving this resulted in a fair bit of refactoring:
    * Refactor the `Llama2ChatFormatter` and `Llama3ChatFormatter` to encapsulate all logic in a single abstract method `encode_dialog_prompt`
    * Remove all formatter-specific logic from the primary generation loop in `def chat`
    * Add the `HFTokenizerChatFormatter`
    * Plumb the ability to use the chat template with `jinja2` through `HFTokenizer`
        * **NOTE**: `jinja2` was already a transitive dependency, so I just formalized it

To get to the bottom of all of this, I also tweaked the logging a bit. There was already a hard-coded logging config call in `cli`, so I just added the ability to parse the `LOG_LEVEL` env var to set it. I also added a fair number of additional log lines and uncommented some that were there but commented out.

**NOTE**: Many of the existing log lines were using f-strings which will cause the string to be interpolated regardless of whether the logger/level are enabled. I switched all of these to use lazy interpolation with percent-encoding so that it's safe to have them uncommented without a performance hit.

Finally, I was getting lost trying to make sure I didn't break anything in the chat templating, so I bit the bullet and added some _basic_ unit tests. They only cover the chat formatting, but they're a place to start. I did not go any further with unit testing, including not adding `pytest` as a dependency or adding any CI steps to invoke the tests. If you're interested, I'd be happy to push on unit testing, but I didn't want to lump that conversation into this PR.

install/requirements.txt
```diff
@@ -9,6 +9,10 @@ gguf
 # Tiktoken tokenizer for Llama 3 and other advanced models
 tiktoken
 
+# Tokenizers and jinja2 for other non-llama models that use HF tokenizers
```
gabe-l-hart, line 12: I added these here, but did not add `pytest` (yet). I think there's a pending conversation about introducing optional dependency sets, so it would make sense to add a `test` or `dev` set at that point, but I didn't want to accidentally carry `pytest` along as a runtime dependency.

tests/conftest.py
```diff
@@ -0,0 +1,12 @@
+"""
+Global pytest config, fixtures, and helpers go here!
+"""
+
+# Standard
+import os
+import sys
+
+# Make sure tests can import torchchat
```
gabe-l-hart, line 9: This would be a lot cleaner if we move to having a `pyproject.toml` or `setup.py` to bundle `torchchat` as a package that could be installed with `pip install -e`.

torchchat/generate.py
```diff
@@ -82,49 +110,83 @@ def encode_message(self, message) -> List[int]:
...
-            tokens.extend(self.encode_message(message))
+            tokens.extend(self._encode_message(message))
         # Add the start of an assistant message for the model to complete.
-        tokens.extend(self.encode_header("assistant"))  # Pass role directly as a string
+        if add_generation_prompt and dialog and dialog[-1]["role"] != "assistant":
+            tokens.extend(self._encode_header("assistant")) # Pass role directly as a string
         return tokens
 
 
-B_INST, E_INST = "[INST]", "[/INST]"
```
gabe-l-hart, line 98: I moved these into the class as members to enforce the encapsulation that they should only be used in the context of this formatter

byjlw:
Thanks @gabe-l-hart 
Yeah a lot of this feedback resonates really well with me, and resolving it has already made it on our H1 roadmap such as making it easy to have model specific templates, adding test infra and guidelines around tests, abstracting and making the code more modular so that there is a specific module for core with well defined APIs that the CLI and API can use. We will also figure out the release strategy and publish two or three specific pip packages. 

Will be able to share the details soon and will have them as RFCs on GH so everyone can comment and contribute.

tests/conftest.py
byjlw: on the list

byjlw (approved):
Longer term we want to implement the tokenizer the drop the dependency on HF tokenizer but this will get things going for now.

Jack-Khuu:
Thanks for all the tests and logging!!!

Looks like the hf_tokenizer got caught in https://github.com/pytorch/torchchat/commit/fff956c5c3a68025025c5d906af80eb44e960ce4, so we'll need a tad of a rebase (we can keep the python code here for ease)

Jack-Khuu:
Good(?) news, the rebase might be a no-op after backing out of the tokenizer migration: https://github.com/pytorch/torchchat/pull/1414

gabe-l-hart:
Ah! I had a half-written comment asking for some guidance on that. Glad I waited. Will try the rebase now.

Jack-Khuu (approved):
This is looking amazing

Can you give `python3 torchchat.py chat <>` a quick check looks like there a little typo? (commented inline)

torchchat/generate.py
```diff
@@ -82,49 +121,83 @@ def encode_message(self, message) -> List[int]:
                         self.tokenizer.encode(content["text"], bos=False, eos=False)
                     )
 
-        tokens.append(self.tokenizer.special_tokens["<|eot_id|>"])
+        tokens.append(self.tokenizer.special_tokens["<|eot_id|>\n"])
```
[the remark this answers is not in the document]
gabe-l-hart: Interesting. It looks like there's some inconsistency in the prompt template documentation on whether or not there should be a `\n` after `<|eot_id|>`:

* Llama 3: https://www.llama.com/docs/model-cards-and-prompt-formats/meta-llama-3/
     * It looks like there _is_ a `\n` after `<|eot_id|>` when it's between `assistant` and `user`, but NOT _always_ when it's between `assistant` and `user` (see the sample under `System prompt and multiple turn...`)
* Llama 3.1: https://www.llama.com/docs/model-cards-and-prompt-formats/llama3_1/#-special-tokens-
    * In almost all samples, there is a `\n` after `<|eot_id|>`, but similar to llama 3, it looks like some times it is omitted for the final header tag when adding the final `assistant` header tag (See sample under `Built in Python based tool calling`)
* Llama 3.2: https://www.llama.com/docs/model-cards-and-prompt-formats/llama3_2/
    * This one is also inconsistent on whether or not there should be a `\n`. In the example under `With the above function definition,...`, it's exactly the _opposite_ pattern of the others where there _isn't_ a `\n` between the `system` and `user`, but then there _is_ between `user` and the generation prompt tag for `assistant` 🤷 
    
All of this is to say, it seems that the documentation is inconsistent. I'm happy to yank it out to keep it consistent with how you had it.

Jack-Khuu:
Kicked off a rebase; fyi has a few broken tests (awaiting pin bump) so those are expected

Jack-Khuu:
Fantastic, thanks for the patience and add!!

Issue #1302, opened 2024-10-15
Out of memory AOTI using llama 3.1 8b on RTX 4090


byjlw:
### 🐛 Describe the bug

I'm getting an out of memory error when trying to use an exported .so file. The .so is 16GB my GPU has 24GB of memory

```
(.venv) warden@Vikander:~/source/torchchat$ python3 torchchat.py generate llama3.1 --dso-path exportedModels/llama3.1.so --prompt "Hello my name is" --device cuda
Warning: checkpoint path ignored because an exported DSO or PTE path specified
Warning: checkpoint path ignored because an exported DSO or PTE path specified
Using device=cuda NVIDIA GeForce RTX 4090
Loading model...
Time to load model: 1.85 seconds
Error: CUDA error: out of memory
Traceback (most recent call last):
  File "/home/warden/source/torchchat/torchchat/cli/builder.py", line 536, in _initialize_model
    model.forward = torch._export.aot_load(
                    ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/warden/source/torchchat/.venv/lib/python3.11/site-packages/torch/_export/__init__.py", line 320, in aot_load
    runner = torch._C._aoti.AOTIModelContainerRunnerCuda(so_path, 1, device)  # type: ignore[assignment, call-arg]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: create_func_( &container_handle_, num_models, device_str.c_str(), cubin_dir.empty() ? nullptr : cubin_dir.c_str()) API call failed at ../torch/csrc/inductor/aoti_runner/model_container_runner.cpp, line 85

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/warden/source/torchchat/torchchat.py", line 88, in <module>
    generate_main(args)
  File "/home/warden/source/torchchat/torchchat/generate.py", line 1215, in main
    gen = Generator(
          ^^^^^^^^^^
  File "/home/warden/source/torchchat/torchchat/generate.py", line 290, in __init__
    self.model = _initialize_model(self.builder_args, self.quantize, self.tokenizer)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1337, opened 2024-11-01
[AOTI] Remove the original model weights in Python deployment

desertfire:
Summary: Fixes https://github.com/pytorch/torchchat/issues/1302. Because AOTI-compiled model contains a copy of model weights, if we also keep the eager model weights, it will be a wast of GPU memory and even triggers OOMs. This PR releases the corresponding eager model weights in the AOTI-Python deployment path.


desertfire:
@Jack-Khuu , is there any known problems with CI? Many of the failures here seem unrelated to my change.

Jack-Khuu:
Yeah, we're checking with the DevInfra folk

Jack-Khuu:
Heads up CI is broken in pt/pt and at higher level

https://github.com/NVIDIA/nvidia-container-toolkit/releases/tag/v1.17.0

byjlw:
@desertfire it should be fixed now. We just need to churn through the tests

mikekgfb:
> Summary: Fixes #1302. Because AOTI-compiled model contains a copy of model weights, if we also keep the eager model weights, it will be a waste of GPU memory and even triggers OOMs. This PR releases the corresponding eager model weights in the AOTI-Python deployment path.

In the long term we might ask whether there's a point in going to all the expense of building the model if we just need the config.  The model build process had a config_only bool argument with the intent of allowing suppression of actual model build (that was never implemented though).  

This is particularly noteworthy (aka expensive because GGUF requires much more than weight mmaps) when building a model from GGUF, as @metascroy pointed out a long time ago when he implemented the GGUF reader.

Jack-Khuu (approved):

PR #1339, opened 2024-11-01
tokenizer was missing an include

byjlw:


Jack-Khuu (approved):

tokenizer/base64.h
```diff
@@ -27,6 +27,7 @@
 #include <cassert>
 #include <string>
 #include <string_view>
+#include <cstdint>
```
larryliu0820, line 30: Nit keep sorted

byjlw: yep

larryliu0820 (approved):

Issue #1298, opened 2024-10-14
RuntimeError: CUDA error: named symbol not found


mikekgfb:
### 🐛 Describe the bug

python torchchat.py generate stories110M --quant torchchat/quant_config/cuda.json --prompt "It was a dark and stormy night, and"

Using device=cuda Tesla T4
Loading model...
Time to load model: 0.73 seconds
Quantizing the model with: {'executor': {'accelerator': 'cuda'}, 'precision': {'dtype': 'bf16'}, 'linear:int4': {'groupsize': 256}}
Time to quantize model: 0.35 seconds
Traceback (most recent call last):
  File "/content/torchchat-1/torchchat.py", line 88, in <module>
    generate_main(args)
  File "/content/torchchat-1/torchchat/generate.py", line 1210, in main
    gen = Generator(
  File "/content/torchchat-1/torchchat/generate.py", line 290, in __init__
    self.model = _initialize_model(self.builder_args, self.quantize, self.tokenizer)
  File "/content/torchchat-1/torchchat/cli/builder.py", line 574, in _initialize_model
    quantize_model(
  File "/content/torchchat-1/torchchat/utils/quantize.py", line 114, in quantize_model
    quantize_(model, int4_weight_only(q_kwargs["groupsize"]))
  File "/usr/local/lib/python3.10/dist-packages/torchao/quantization/quant_api.py", line 462, in quantize_
    _replace_with_custom_fn_if_matches_filter(
  File "/usr/local/lib/python3.10/dist-packages/torchao/quantization/quant_api.py", line 202, in _replace_with_custom_fn_if_matches_filter
    new_child = _replace_with_custom_fn_if_matches_filter(
  File "/usr/local/lib/python3.10/dist-packages/torchao/quantization/quant_api.py", line 202, in _replace_with_custom_fn_if_matches_filter
    new_child = _replace_with_custom_fn_if_matches_filter(
  File "/usr/local/lib/python3.10/dist-packages/torchao/quantization/quant_api.py", line 202, in _replace_with_custom_fn_if_matches_filter
    new_child = _replace_with_custom_fn_if_matches_filter(
  [Previous line repeated 2 more times]
  File "/usr/local/lib/python3.10/dist-packages/torchao/quantization/quant_api.py", line 198, in _replace_with_custom_fn_if_matches_filter
    model = replacement_fn(model)
  File "/usr/local/lib/python3.10/dist-packages/torchao/quantization/quant_api.py", line 392, in insert_subclass
[...]

PR #1344, opened 2024-11-05
recognize and issue error if GPU does not support bf16

mikekgfb:
Address pytorch/torchchat#1298 which causes models to fail on T4 (and other pre-9.0 arch level GPUs) by selecting an alternate dtype when possible, and issue an error otherwise

### DIFF
--- a/torchchat/utils/build_utils.py
+++ b/torchchat/utils/build_utils.py
@@ -148,22 +148,51 @@ def get_precision():
 ###               dtype name to torch.dtype mapping                    ###
 
 
+def get_cuda_architecture(device=None):
+    device_str = get_device_str(device)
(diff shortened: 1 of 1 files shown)


mikekgfb:
Because we emulate BF16 on pre-V9.0 CUDA architecture, this test is overly restrictive.  In a nutshell, the only problem are a small set of functions, like torchao's linear:int4 operator (xref: https://github.com/pytorch/ao/issues/1110), that don't emulate FP16.  Given the general posture of PyTorch, emulation for those operators, rather than issuing an error, would be the way to go.

mikekgfb:
Closing, as this PR implements an overly restrictive check given the emulation of BF16 for older architectures.

PR #1346, opened 2024-11-06
fix: allow multiple weight mapping files for mistral

leseb:
42f18114 fix: allow multiple weight mapping files for mistral

commit 42f181148b5a077adde5132f66575c64807ac260
Author: Sébastien Han <seb@redhat.com>
Date:   Wed Nov 6 15:20:28 2024 +0100

    fix: allow multiple weight mapping files for mistral
    
    Downloading a Mistral model fails because it includes multiple weight
    mapping files. The regression was introduced in commit
    `766bee9f4a1fcb187fae543a525495d3ff482097`. I'm unclear on the original
    intent, but perhaps the exception was meant to apply only to Granite
    models. This isn’t an ideal fix, but it does enable Mistral to be
    downloaded and used for chat.
    
    Signed-off-by: Sébastien Han <seb@redhat.com>

### DIFF
--- a/torchchat/cli/convert_hf_checkpoint.py
+++ b/torchchat/cli/convert_hf_checkpoint.py
@@ -43,7 +43,8 @@ def convert_hf_checkpoint(
 
     # Load the json file containing weight mapping
     model_map_json_matches = [Path(m) for m in glob.glob(str(model_dir / "*.index.json"))]
-    assert len(model_map_json_matches) <= 1, "Found multiple weight mapping files"
+    if "mistral" not in model_name:
+        assert len(model_map_json_matches) <= 1, "Found multiple weight mapping files"
     if len(model_map_json_matches):
         model_map_json = model_map_json_matches[0]
     else:


leseb:
CLA is in-progress.

Jack-Khuu:
Thanks for the fix  @leseb, kicked off the CI

gabe-l-hart:
@leseb Good find! The change I made in [766bee](https://github.com/pytorch/torchchat/commit/766bee9f4a1fcb187fae543a525495d3ff482097) was simply to extend from using a single hard-coded name to allow for a single non-hard coded name, so the limitation of a single weight mapping was there before. With this change, it will not assert, but it will still only ready the weight mapping from the first file found. In the case of `mistral`, are the multiple weight files redundant? I suppose we could do some kind of check and/or merge of the multiple files if the mapping is split across multiple files, though I'm not clear if that's actually a valid file layout.

gabe-l-hart:
Ah, I see in [the files](https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.1/tree/main) that the multiple `index` files are mapped to the multiple different weight format sets (`.safetensors` vs `.bin`). I think the right fix would be to somehow line them up based on the `prefer_safetensors` value for the model.

gabe-l-hart:
I've added a few more commits on top of @leseb's: https://github.com/leseb/torchchat/pull/1

leseb:
> I've added a few more commits on top of @leseb's: [leseb#1](https://github.com/leseb/torchchat/pull/1). I'm not sure whether the best course is to merge my commits into his branch to update the pointer for this PR or to just update this branch's pointer to use my fork. I'm open to either!

@gabe-l-hart Thanks for taking the time to come up with a better approach. At this stage, I believe it makes more sense that you take over since there is nothing left from my initial change. Thanks!

Issue #1347, opened 2024-11-06
Slow CLI --help (and other commands)


gabe-l-hart:
### 🐛 Describe the bug

When running the `torchchat.py` CLI, even with `--help`, it can take several seconds. This is caused by the eager imports of expensive packages (particularly `torch`) which are unnecessary for some of the functionality, but are imported regardless, even before constructing the `argparse` parser.

The proposed fix is to move all of the expensive imports so that they happen just-in-time. This would enable `--help` (and `list`/`where`) to run without needing to process the expensive imports first.

### Versions

Collecting environment information...
PyTorch version: 2.6.0.dev20241002
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 14.5 (arm64)
GCC version: Could not collect
Clang version: 15.0.0 (clang-1500.3.9.4)
CMake version: version 3.30.4
Libc version: N/A

Python version: 3.10.15 | packaged by conda-forge | (main, Sep 30 2024, 17:48:38) [Clang 17.0.6 ] (64-bit runtime)
Python platform: macOS-14.5-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M3 Max

Versions of relevant libraries:
[pip3] numpy==1.26.4
[pip3] torch==2.6.0.dev20241002
[pip3] torchao==0.5.0
[pip3] torchtune==0.4.0.dev20241010+cpu
[pip3] torchvision==0.20.0.dev20241002
[conda] numpy                     1.26.4                   pypi_0    pypi
[conda] torch                     2.6.0.dev20241002          pypi_0    pypi
[conda] torchao                   0.5.0                    pypi_0    pypi
[conda] torchtune                 0.4.0.dev20241010+cpu          pypi_0    pypi
[conda] torchvision               0.20.0.dev20241002          pypi_0    pypi

PR #1349, opened 2024-11-06
Fast non-model CLI commands

gabe-l-hart:
## Description

Closes: https://github.com/pytorch/torchchat/issues/1347

Import torch lazily in all places used by the CLI that don't need a model

These changes add a little complexity with the lazy and local imports, but they also greatly improve the CLI's response for `--help`, `list`, and `where`.

**Changes:**

* Move `import torch` into function bodies that need them
* Use `importlib.metadata.version` to check the torch version rather than torch.__version__
* Switch from using torch.inference_mode as a decorator to using it as a context manager.
  * I also removed it from convert_hf_checkpoint_to_tune since that does not use torch at all
* In build_utils, wrap the dtype values in lambdas so they're lazily fetched.

### DIFF
--- a/torchchat/cli/cli.py
+++ b/torchchat/cli/cli.py
@@ -5,20 +5,16 @@
 # LICENSE file in the root directory of this source tree.
 
 import argparse
+import importlib.metadata
 import json
 import logging
 import os
 import sys
 from pathlib import Path
 
-import torch
-
-from torchchat.cli.download import download_and_convert, is_model_downloaded
-
 from torchchat.utils.build_utils import (
     allowable_dtype_names,
     allowable_params_table,
-    get_device_str,
 )
 
 logging.basicConfig(level=logging.INFO, format="%(message)s")
@@ -42,6 +38,9 @@ KNOWN_VERBS = GENERATION_VERBS + ["eval", "export"] + INVENTORY_VERBS
 
 # Handle CLI arguments that are common to a majority of subcommands.
 def check_args(args, verb: str) -> None:
+    # Local import to avoid unnecessary expensive imports
+    from torchchat.cli.download import download_and_convert, is_model_downloaded
+
     # Handle model download. Skip this for download, since it has slightly
     # different semantics.
     if (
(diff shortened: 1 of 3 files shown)


torchchat/cli/cli.py
```diff
@@ -498,9 +497,10 @@ def _add_speculative_execution_args(parser) -> None:
 
 
 def arg_init(args):
-    if not (torch.__version__ > "2.3"):
+    torch_version = importlib.metadata.version("torch")
+    if not torch_version or (torch_version <= "2.3"):
         raise RuntimeError(
-            f"You are using PyTorch {torch.__version__}. At this time, torchchat uses the latest PyTorch technology with high-performance kernels only available in PyTorch nightly until the PyTorch 2.4 release"
+            f"You are using PyTorch {torch_version}. At this time, torchchat uses the latest PyTorch technology with high-performance kernels only available in PyTorch nightly until the PyTorch 2.4 release"
```
mikekgfb, line 503: We might defer that to commands that actually use torch,  and then we can just use torch.__version__ ?  No point in raising a version issue when all we do is run help or similar non-model cli commands? 

Wdyt @Jack-Khuu

byjlw: I like having it here, since we can catch it at the very beginning as opposed to having version checking in every file. 

Doing it this way will speed up the interface by a lot. Right now --help and --command is super slow. It also bogs down CI.

byjlw (approved):

torchchat/cli/cli.py
gabe-l-hart: I could see it either way, but I think I agree with @byjlw. I don't think the non-running commands are valuable in-and-of-themselves, so I think the argument would be that this is really a property of the tool (there is no `torchchat` without `torch`) and doing the lazy imports is just a speed optimization.

If there were standalone commands that didn't need `torch` (i.e. if people start using `download` just to fetch models, but then run them with something else), then I could see making this a soft requirement and deferring it to the commands that need it.

Jack-Khuu: I actually added what @mikekgfb is suggesting a while back, so this check being in `arg_init` actually does what all 3 y'all suggest (check only if needed, and check in one place)

https://github.com/pytorch/torchchat/blob/4a7dab8cfb7111aa2323ad840cda68d65b81e86f/torchchat.py#L62-L63

`arg_init` is only called when it is pertinent

Jack-Khuu (approved):
Thanks for adding this @gabe-l-hart!!

It's been on my list for a while

PR #1355, opened 2024-11-07
[WIP] Generate base class for better integration of distributed inference

mreso:
This PR introduces a new Generator base class to better integrate distributed inference with the already present infra

### DIFF
--- a/torchchat/generate.py
+++ b/torchchat/generate.py
@@ -17,6 +17,7 @@ from io import BytesIO
 from os import PathLike
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
+from typing_extensions import override
 
 import torch
 import torch._dynamo.config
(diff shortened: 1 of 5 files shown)


torchchat/generate.py
```diff
@@ -3,218 +3,35 @@
...
-            compile_prefill=args.compile_prefill,
-            speculate_k=args.speculate_k,
-            sequential_prefill=sequential_prefill,
-            max_autotune=args.max_autotune,
-            is_torchtune_model=args.model and args.model.endswith("tune"),
-        )
-
-
-class Generator:
+class SingleGPUGenerator(Generator):
```
mikekgfb, line 34: I assume this name was chosen as contrast to distributed multi-GPU? There's nothing specific to GPU here, and the same generator is also used for ET models, CPU and accelerators.  How about `LocalGenerator` for the base case that's just running on the present node's resources, whatever they may be?

mikekgfb (changes requested):
Rename SingleGpuGenerator -> LocalGenerator, or some other name that reflects the broader use of this class, please?

torchchat/generate.py
mreso: Good point! Thanks @mikekgfb

mreso: Fixed

mreso:
Hi @Jack-Khuu would be great to get your opinion on this PR as well because it does some more aggressive refactoring to integrate distributed inference. Its only a step into the direction to keep PR size small. The final goal is to have LocalGenerator and DistributedGenerator use a unified interface. Then we can reuse the code base for preparing the data etc by sharing the chat() method which would eventually also be raise into the base class as well. [This branch](https://github.com/mreso/torchchat/tree/refactor/move_tokenizer_out_of_worker) contains the next PR where I aligned the generate() interface. Next step is to refactor chat() to support both Local- and DistributedGenerator.

Jack-Khuu (approved):

Jack-Khuu:
CI is fixed

The changes with LocalGenerator look intuitive to me and I see that generate is going to be pushed up a layer into Generate in your next PR which is great.

Left one question about `gen_model_input`, but once that's resolved we're gtg

torchchat/utils/generator.py
```diff
@@ -0,0 +1,399 @@
...
+        """
+        raise NotImplementedError()
+
+    def _gen_model_input(
+        self,
+        prompt: Union[str | List[Any]],
+        image_prompts: Optional[List[str | Image.Image]] = None,
+        max_new_tokens: Optional[int] = None,
+        max_seq_len: Optional[int] = 2048,
+    ) -> Tuple[torch.Tensor, Optional[Dict[str, Any]]]:
```
Jack-Khuu, line 246: Something similar to generator was something we were looking into for Q4/H1, so I'm glad to see the intial work here lining up with my mental model.

Is `_gen_model_input` used in the `DistributedGenerator`? If not let's keep this in the `LocalGenerator` so that `Generator` is more succinct.

Jack-Khuu: We can be more aggressive with leaving things in LocalGenerator and Generator being thin, but that's something we can do later (but gen_model_input should be moved unless there's a need)

mreso: Thanks for reviewing and feedback. The "copy and pastability" design aspect of TorchChat got to me through the recent discussions. Regarding _gen_model_input, I was planning to only move things into Generator that I could reuse in distributed. In my dev branch I actually pulled [chat into the base class](https://github.com/mreso/torchchat/blob/refactor/uplift_chat_into_base_generator_class/torchchat/utils/generator.py#L259
) to reuse it so _gen_model_input would need to live in Generator as well. I think in the end you could end up with only load_model, decode_one_token and decode_n_tokens and some model properties in the Local/DistributedGenerator which in some sense would make it "copy and pastability" again as everything lives in Generator.

mreso:
Lets hold on merging this, and let me look into a more "copy and pastabile" approach. As we'll end up with only model specifics in the Local/DistributedGenerator we might get away with just touching the model level.

mreso:
Closing this in favor of #1381

Issue #1278, opened 2024-10-06
AOTI Export ignores user --device flag - expected behavior? 


vmpuri:
### 🐛 Describe the bug

Hi all, 

I ran into some confusion when trying to export llama3 on my system. I have a small graphics card (8GB VRAM on an AMD GPU) but a decent amount of RAM (24GB). Obviously, the model won't fit on my GPU un-quantized but it should fit into my RAM + swap.

I tried running:
```
python3 torchchat.py export llama3 --output-dso-path exportedModels/llama3.so --quantize torchchat/quant_config/desktop.json  --device cpu
```

However, I ran into multiple HIP OOM errors (basically equivalent to CUDA). Why would we try to allocate CUDA memory if the target device is CPU?

On further inspection, during export, the device is replaced with whatever is present in the quantize config:
In `cli.py`
https://github.com/pytorch/torchchat/blob/b21715835ab9f61e23dbcf32795b0c0a2d654908/torchchat/cli/cli.py#L491C10-L494C1
```
args.device = get_device_str(
    args.quantize.get("executor", {}).get("accelerator", args.device)
)
```

In this case, the device in `desktop.json` is "fast". The `get_device_str` function replaces this with "cuda" simply based on `torch.cuda.is_available` without consulting the flag I passed in. 

## Other cases
Doing a quick grep of the repo, I only found one other case in `generate.py` where `torch.cuda.is_available()` is consulted for monitoring memory usage. We should be careful switching based simply on `torch.cuda.is_available()` and make sure to pin to the user's request if we're using ambiguous devices like "fast".

Another small issue - since I use AMD GPU, the default `install/install_requirements.sh` will download the CPU only version instead of the ROCm version of PyTorch. To use my GPU, I have to re-run the torch installation manually. Luckily, it's quite easy to find this command at https://pytorch.org/get-started/locally/ . Should be straightforward to check of ROCm is available on the system during this script - we can just run `rocminfo` & check if the command is available.

### Versions

```
wget https://raw.githubusercontent.com/pytorch/pytorch/main/torch/utils/collect_env.py
# For security purposes, please check the contents of collect_env.py before running it.
python collect_env.py
--2024-10-06 12:03:44--  https://raw.githubusercontent.com/pytorch/pytorch/main/torch/utils/collect_env.py
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 2606:50c0:8001::154, 2606:50c0:8002::154, 2606:50c0:8003::154, ...
[...]

PR #1359, opened 2024-11-08
Update cli.py to make --device/--dtype pre-empt quantize dict-specified values

mikekgfb:
Users may expect that cli parameters override the JSON, as per #1278.   Confirm the inverted behavior suggested by #1278 is what we want.  Also, if we do, then for sanity, we should apply the same priority scheme to both dtype and device values.

Invert logic - case split: 

1 - if none (no value) is specified, use value specified in quantize dict, if present; else 
2 - if value is specified, override the respective handler if present.


Jack-Khuu (approved):
Beautiful, thanks again @mikekgfb

PR #1368, opened 2024-11-12
Minor typo + Update install_requirements.sh to support python 3.10 >=

Jack-Khuu:
Minor fixes to install_requirement (undefined variable) + Loosen the Python requirements to 3.10.0+
* We can case out specific Python version as needed, but no need to restrict preemptively

---
Fails with 3.8
```
./install/install_requirements.sh
Using python executable: python3
Python version must be greater than 3.10.x. Detected version: 3.8
```

Works with 3.10.0

### DIFF
--- a/install/install_requirements.sh
+++ b/install/install_requirements.sh
@@ -19,14 +19,14 @@ fi
 echo "Using python executable: $PYTHON_EXECUTABLE"
 
 PYTHON_SYS_VERSION="$($PYTHON_EXECUTABLE -c "import sys; print(f'{sys.version_info.major}.{sys.version_info.minor}')")"
-# Check python version. Expect 3.10.x or 3.11.x
+# Check python version. Expect 3.10.x to 3.12.x. Note 3.13 may work, but no guarantees
 if ! $PYTHON_EXECUTABLE -c "
 import sys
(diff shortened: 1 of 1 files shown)


Jack-Khuu:
Going to use this PR to address some other related instrall_req bugs and open python to 3.10+

leseb (approved):

install/install_requirements.sh
```diff
@@ -14,19 +14,21 @@ then
...
     PYTHON_EXECUTABLE=python3
+  else
+    PYTHON_EXECUTABLE=python
   fi
 fi
 echo "Using python executable: $PYTHON_EXECUTABLE"
 
 PYTHON_SYS_VERSION="$($PYTHON_EXECUTABLE -c "import sys; print(f'{sys.version_info.major}.{sys.version_info.minor}')")"
-# Check python version. Expect 3.10.x or 3.11.x
+# Check python version. Expect greater than 3.10.x
```
Gasoonjia, line 24: word and code below seem mismatch -- here we say greater than 3.10.x, but the code below is for greater than or equlal to 3.10

Jack-Khuu: Updated wording

PR #1375, opened 2024-11-14
fix: do not print perf stat when NaN

leseb:
b4153260 fix: do not print perf stat when NaN

commit b41532608d0f1648285d7794eda4331b1cfb297f
Author: Sébastien Han <seb@redhat.com>
Date:   Thu Nov 14 11:04:47 2024 +0100

    fix: do not print perf stat when NaN
    
    If the chat is exited or interrupted it will still print the stats with
    NaN values which is unnecessary.
    
    Signed-off-by: Sébastien Han <seb@redhat.com>

### DIFF
--- a/torchchat/generate.py
+++ b/torchchat/generate.py
@@ -1189,12 +1189,27 @@ with {'sequential' if generator_args.sequential_prefill else 'parallel'} prefill
                 f"Mean Accepted: {sum([idx * i for idx, i in enumerate(counts_aggregated)])/sum(counts_aggregated)}"
             )
 
-        print(
-            f"\n      Average tokens/sec (total): {torch.mean(torch.tensor(aggregate_metrics['tokens_per_sec'])).item():.2f} \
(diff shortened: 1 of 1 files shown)


Jack-Khuu:
PR looks good, but can you share an example of when this would get triggered (i.e. when are we seeing NaN via manually kill)?

leseb:
> PR looks good, but can you share an example of when this would get triggered (i.e. when are we seeing NaN via manually kill)?

```
python3.10 torchchat.py chat llama3.1 
NumExpr defaulting to 12 threads.
PyTorch version 2.6.0.dev20241002 available.
Using device=mps 
Loading model...
Time to load model: 15.06 seconds
Starting Interactive Chat
Entering Chat Mode. Will continue chatting back and forth with the language model until the models max context length of 8192 tokens is hit or until the user says /bye
Do you want to enter a system prompt? Enter y for yes and anything else for no. 

User: /bye
Exiting Chat.


      Average tokens/sec (total): nan                 
Average tokens/sec (first token): nan                 
Average tokens/sec (next tokens): nan 
```

Jack-Khuu (approved):

PR #1378, opened 2024-11-15
Remove tokens per sec in aggregate_metrics when jit_compile

yanbing-j:
This PR is to remove tokens per sec in aggregate_metrics when jit_compile, to make the average of tokens per sec more precise. The tokens per sec is too low when jit_compile, which should not be included in aggregate_metrics.

### DIFF
--- a/torchchat/generate.py
+++ b/torchchat/generate.py
@@ -1149,9 +1149,11 @@ class Generator:
                 print(
                     f"just-in-time compilation time (incl run time): {compilation_time:.2} seconds"
                 )
-            aggregate_metrics["tokens_per_sec"].append(tokens_sec)
-            aggregate_metrics["first_token_per_sec"].append(first_token_sec)
-            aggregate_metrics["next_tokens_per_sec"].append(next_tokens_sec)
+            else:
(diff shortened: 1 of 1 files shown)


yanbing-j:
Hi @Jack-Khuu, could you please take a look? Thanks!

Jack-Khuu:
Thanks for adding. Looks like the compile numbers got added back in (outside of the if) with https://github.com/pytorch/torchchat/commit/77774d2345dee150d7bdc2dbd22529cbde388ed7 due to Nan's

I rebased your PR since https://github.com/pytorch/torchchat/commit/d7b681a0b8745874d11db2246a7eefbf23b07a5a added a Nan check that might actually make the old Nan problem a non-issue

yanbing-j:
> the averages were calculated with the compile sample excluded

Updated. Please take a look again.

Jack-Khuu:
Thanks for the comment

Just one last nit: Let's only print the "excluding compile in calculations" message if the user is actually compiling

Once that's ready give me a ping and I'll merge this in

yanbing-j:
@Jack-Khuu Thanks for the merging!

Issue #1376, opened 2024-11-14
[RFC] Integration of Distributed Inference into TorchChat


mreso:
### 🚀 The feature, motivation and pitch

**Overview**
The goal of this RFC is to discuss the integration of distributed inference into TorchChat. Distributed inference leverages tensor parallelism or pipeline parallelism, or a combination of both to support larger model size which do not fit on a single accelerator. Through parallelization each model shard runs in its own worker process. The processes can either be spawned on the script level (e.g. via torchrun) or from within the main script. For online use cases like chat/server the processes need to coordinate fetching and sharing the user input depending on at which point the processes get spawned. Synchronization points between the processes should be minimized for optimal performance.
The design goals of the integration are:
* Support all CLI features of TorchChat (generate, chat, server)
* Minimize code duplication
* Maintain TorchChat's copy/pastebility

### Alternatives

**Option 1: Integrate at Model Level**
While the usage of a tensor parallel model in PyTorchis is very much transparent, the current pipeline parallel API differs significantly from the usage of a local model. This option hides the distributed inference from the Generator class by introducing the distributed inference inside a torchchat.model.Model derivative. The DistributedModel(torchchat.model.Model) class would implement methods like __call__() and forward() and handle distribution to the worker processes inside. 
* Pros: 
   * Code reuse high
   * Transparent use of distributed model
   * Virtually no changes in main Generator and OpenAiApiGenerator necessary
* Cons:
   * In this scenario, sampling happens in the main script and thus the return value of the model (logits) need to be transferred between processes (i.e. moved to shared GPU memory)
   * As the Generator is unaware of the parallelism the subprocesses would need to be spawned inside the model itself which is kind of ugly

**Option 2: Abstract Base Class for Generator**
Introduce a base class Generator which contains the common portions of the implementation generation process like getting and preparing input from the user. LocalGenerator and DistributedGenerator get introduced to handle specifics. The split between base and derivatives can be made at multiple levels, specifically High:Generator.generate, Mid:Generator.decode_n_tokens/prefill, Low: Generator.decode_one_token/prefill
* Pros:
   * Introduces abstraction in the generation process
   * High code reuse
   * Subprocess creation for parallel workers can be on main script level
   * Added complexity stays mostly separate from local generation
* Cons:
   * Splitting up the Generator from main generate.py file will hurt copy/pastebility
   * OpenAiApiGenerator (currently inherits from Generator) will require additional changes to work with distributed inference

[...]

PR #1327, opened 2024-10-24
Integrate distributed inference into torchchat cli

mreso:
This PR is the first step towards integrating distributed inference into the torchchat CLI.
Currently only `torchchat.py generate` is supported.

Test:
`python torchchat.py generate llama3.1 --distributed --max-new-tokens 40 --prompt "What is Snow?"`
Output:
```
Launching distributed inference ...
10-24 12:48:32.449 - dist_run:304 - Worker started: rank=0, world_size=2
10-24 12:48:32.449 - dist_run:307 -  GPU capacity: NVIDIA PG509-210 (None) with 79.15GiB memory
10-24 12:48:32.449 - dist_run:310 - Using model weights from meta-llama/Meta-Llama-3.1-8B-Instruct and dtype torch.bfloat16
known configs: ['13B', '30B', '34B', '70B', '7B', 'CodeLlama-7b-Python-hf', 'Llama-3.2-11B-Vision', 'Llama-Guard-3-1B-INT4', 'Llama-Guard-3-1B', 'Meta-Llama-3-70B', 'Meta-Llama-3-8B', 'Meta-Llama-3.1-70B-Tune', 'Meta-Llama-3.1-70B', 'Meta-Llama-3.1-8B-Tune', 'Meta-Llama-3.1-8B', 'Meta-Llama-3.2-1B', 'Meta-Llama-3.2-3B', 'Mistral-7B', 'llava-1.5', 'stories110M'
, 'stories15M', 'stories42M']
10-24 12:48:32.450 - dist_run:316 - Transformer Config: TransformerArgs(block_size=131072, vocab_size=128256, n_layers=32, n_heads=32, dim=4096, hidden_dim=14336, n_local_heads=8, head_dim=128, rope_base=500000.0, norm_eps=1e-05, multiple_of=1024, ffn_dim_multiplier=1.3, use_tiktoken=True, max_seq_length=8192, rope_scaling={'factor': 8.0, 'low_freq_factor': 1.
0, 'high_freq_factor': 4.0, 'original_max_position_embeddings': 8192}, n_stages=1, stage_idx=0, attention_bias=False, feed_forward_bias=False, tie_word_embeddings=False)
10-24 12:48:32.693 - dist_run:112 - using tokenizer = tokenizer.tiktoken.Tokenizer
10-24 12:48:32.693 - dist_run:321 - Using cache precision torch.bfloat16
[...]

PR #1381, opened 2024-11-16
Integrate distributed inference with chat/server

mreso:
This PR integrates distributed inference with the CLI options chat and server and implements Option 2b of #1376.

To test run on machine with 4 gpus:

```
python  torchchat.py chat llama3.1  --max-new-tokens 10  --distributed --tp 2 --pp 2
python  torchchat.py server  llama3.1  --max-new-tokens 10  --distributed --tp 2 --pp 2
```


torchchat/cli/builder.py
```diff
@@ -598,6 +606,117 @@ def do_nothing(max_batch_size, max_seq_length):
...
+    elif builder_args.distributed:
+        # Using params_table to identify the model to load, for example "Meta-Llama-3.1-8B".
+        #TODO This is a hacky way to please the distributed loading api and needs to be replaced
+        NAME_TO_DISTRIBUTION = {
+            "Meta-Llama-3-8B": "meta-llama/Meta-Llama-3-8B-Instruct", 
+            "Meta-Llama-3.1-8B": "meta-llama/Meta-Llama-3.1-8B-Instruct",
+            "Meta-Llama-3-70B": "meta-llama/Meta-Llama-3-70B-Instruct",
+            "Meta-Llama-3.1-70B": "meta-llama/Meta-Llama-3.1-70B-Instruct",
+
+        }
```
mikekgfb, line 618: What's the gap to using the models described in `model_config/models.json`.  

Definitely should not be part of the present PR, but I think as a north star, it would be desirable to grab the same models (and download and mgmt infra etc) for non-distributed and distributed.

mreso: Thanks for the comment @mikekgfb. The gap wasn't that big, the args.model just wasn't accessible at that point and I wanted to take a deeper look to fix it right. Removed the cruel hack and I now save the distribution_path when creating the builder_arg. Still not sure if this is the intended way of locating the checkpoint though.

mikekgfb: Not sure what you're trying to do.  If it's just that the dict/json data structure describing the mapping isn't in scope, maybe what you want is some methods that give you the relevant info?

Also, these seem to be mapping of short names to HF network paths - should we not have a way to [ick them from the local filesystem (since the torchchat cli already manages download and all that).  Oh, and if the answer is "we have bigger fish to fry, hooking this up is not highest priority" I will wholeheartedly agree.  This is more about understanding the context of this PR.  

Where I'm lacking the context is how you go from all the weights being available locally on a node to reading those weights on another node?  And maybe that's why you prefer to straight up pick the files from HF?  (Although local distribution from an already downloaded set of weights probs has higher bandwidth?)  Again, there's much bigger fish to fry, and I think this PR is a good step in the direction of frying those fish ;)

torchchat/usages/openai_api.py
```diff
@@ -486,6 +486,20 @@ def _callback(self, x, *, buffer, done_generating):
         pass
 
 
+def create_openai_api_generator(distributed):
```
Jack-Khuu, line 489: Nit: typehint

torchchat/usages/openai_api.py
```diff
@@ -486,6 +486,20 @@ def _callback(self, x, *, buffer, done_generating):
...
+    """
+    Factory method to create an OpenAiApiGenerator
+    """
+
+    if distributed:
+        # Base class order matters to make sure OpenAiApiGeneratorMixin overrides methods in DistributedGenerator and Generator
+        return type('OpenAiApiGenerator', (OpenAiApiGeneratorMixin, DistributedGenerator), {})
```
Jack-Khuu, line 496: Nit: we can case or ternary just on the Generator class since the rest is the same

torchchat/usages/server.py
```diff
@@ -4,38 +4,93 @@
...
+    procs = []
+    if builder_args.distributed:
+        world_size = builder_args.tp * builder_args.pp
+        mp_context = mp.get_context('spawn')
+        queue = mp_context.Queue()
+    else:
+        world_size = 1
+        queue = None
+
+
```
Jack-Khuu, line 78: Both if statements check for is_distributed. Can combine?

torchchat/usages/server.py
```diff
@@ -4,38 +4,93 @@
...
 
-    gen: OpenAiApiGenerator = initialize_generator(args)
+    builder_args = BuilderArgs.from_args(args)
+    procs = []
+    if builder_args.distributed:
+        world_size = builder_args.tp * builder_args.pp
+        mp_context = mp.get_context('spawn')
+        queue = mp_context.Queue()
+    else:
+        world_size = 1
```
Jack-Khuu, line 75: Is the else case necessary? I don’t think these fields are used if not distributed

Jack-Khuu (commented):
Partial review so far, still need to dig through generate 

cc: @Gasoonjia as secondary reviewer

torchchat/cli/builder.py
mreso: Oh I see, let me try to fill in some context for this PR. Previously, the distributed inference solution lived completely separate in its own script (within the torchchat repo but) separate from torchchat.py. Distributed [provides its own utils](https://github.com/pytorch/torchchat/blob/edc2cfbd81b54f60a31239588007bba619f91fd7/torchchat/distributed/checkpoint_utils.py) to load either hf or torchchat weights (where the torchchat part is currently broken IIRC). In a previous PR (#1327), I enabled the usage of `torchchat.py generate` with a distributed model. This PR only progresses the integration into the cli by enabling chat/server but stops short from replacing the weight loading mechanics which are still custom to distributed.

So, yes, for this PR I was only looking for a quick and dirty way to map arg.model_name -> "huggingface distribution str" (without actually having model_name at hand) to load the weights from the [hf cache](https://github.com/pytorch/torchchat/blob/edc2cfbd81b54f60a31239588007bba619f91fd7/torchchat/distributed/checkpoint_utils.py#L67). I now modified the PR to use the information provided in the model_config/models.json as you suggested. In a next PR we should then alter `torchchat/distributed/checkpoint_utils.py` to leverage torchchat infra like (e.g. builder_args) to locate and access the files.

torchchat/generate.py
```diff
@@ -1246,20 +1557,20 @@ def main(args):
...
+        ctx = mp.get_context('spawn')
+        with futures.ProcessPoolExecutor(max_workers=world_size-1, mp_context=ctx) as executor:
+            for i in range(1,world_size):
+                fn = partial(run_generator, args, i)
+                executor.submit(setup_env, world_size, i, fn)
+            #Starting rank 0
+            fn = partial(run_generator, args, 0)
+            setup_env(world_size, 0, fn)
+    else:
+        run_generator(args)
```
Jack-Khuu, line 1576: Shuold `setup_env` also be triggering the generator?

Maybe it's just a naming change (run_with_setup~?), but setup_env doesn't make me think that the fn is called

torchchat/generate.py
```diff
@@ -39,11 +43,16 @@
     BuilderArgs,
     TokenizerArgs,
 )
-from torchchat.distributed.generate import DistributedGenerator
+from torchchat.distributed.utils import (
+    Color as color,
+    setup_env,
```
Jack-Khuu, line 48: nit: `setup_env` is pretty generic, let's alias it so it's obviously for a distributed context

torchchat/usages/server.py
```diff
@@ -4,38 +4,90 @@
...
-    gen: OpenAiApiGenerator = initialize_generator(args)
+    builder_args = BuilderArgs.from_args(args)
+    procs = []
+    queue = None
+    if builder_args.distributed:
+        world_size = builder_args.tp * builder_args.pp
+        mp_context = mp.get_context('spawn')
+        queue = mp_context.Queue()
+    
+    if builder_args.distributed:
```
Jack-Khuu, line 76: Identical conditional check

torchchat/generate.py
```diff
@@ -1217,21 +1224,325 @@ def callback(x, *, done_generating=False):
...
+        decoder = ScheduleGPipe(decode_stage, 1)
+
+        return decoder
+
+    def prefill(
+        self,
+        model: Model,
+        x: torch.Tensor,
+        input_pos: torch.Tensor,
+        batch: Optional[Dict[str, Any]] = None,  # Inputs for multimodal models
```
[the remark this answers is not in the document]
mreso: No, its currently text only for distributed.

torchchat/generate.py
mreso: Yes, valid point. I've renamed it to run_in_dist_env to reflect that it runs the target as well as connecting it to distributed.

Jack-Khuu (approved):

mikekgfb:
> This PR integrates distributed inference with the CLI options chat and server and implements Option 2b of #1376.
> 
> To test run on machine with 4 gpus:
> 
> ```
> python  torchchat.py chat llama3.1  --max-new-tokens 10  --distributed --tp 2 --pp 2
> python  torchchat.py server  llama3.1  --max-new-tokens 10  --distributed --tp 2 --pp 2
> ```

Should `torchchat generate` work as well, code would suggest that we can do like so:
```
python  torchchat.py generate llama3.1  --max-new-tokens 10  --distributed --tp 2 --pp 2
```

mreso:
@mikekgfb Yes, distributed generate was already working before this PR, thats why I did not list it separately.

PR #1384, opened 2024-11-18
Update README.md to run and query server during test

mikekgfb:
1 - Run server:
      1a - in background
      1b - capture server_pid

2 - enable query using curl

3 - shutdown server with server pid captured in server_pid


Jack-Khuu:
Nice fix

Jack-Khuu (approved):

mikekgfb:
@Jack-Khuu maybe this is failing because we're hitting the 30 minute limit on tests?  There's not much we can do here because we're not even out of installation  (it appears this test gets killed during `+ ./install/install_requirements.sh` upon closer review).
In particular, splitting the test won't do anything either :(

Updated timeout to see if this may complete in 45 minutes.

Jack-Khuu:
Working on a pin bump to fix the CI noise

PR #1390, opened 2024-11-20
Changing the referenced AAR so that it uses the AAR from the docs

infil00p:
- Trivial edit to ignore the .idea directory when opening torchchat with Android Studio.
- Renamed the AAR referenced so that it's consistent with the instructions provided in the README for deploying on Android

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,10 @@ runner-et/cmake-out/*
 runner-aoti/cmake-out/*
 cmake-out/
 
+# Example project Android Studio ignore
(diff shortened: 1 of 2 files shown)


Jack-Khuu:
Thanks for the fix @infil00p!!

I'm going to give @kirklandsign (Android) a chance to sanity check/merge, but it lgtm. I'll merge if he doesn't get around to it

Jack-Khuu (approved):

kirklandsign:
Thank you! Good catch!

kirklandsign (approved):

Issue #1385, opened 2024-11-19
Update dead link in https://github.com/pytorch/torchchat/blob/main/docs/quantization.md


yanbing-j:
### 🐛 Describe the bug

There is a dead link https://github.com/pytorch/torchchat/blob/main/torchchat/utils/quantize.py#L1260-L1266 in https://github.com/pytorch/torchchat/blob/main/docs/quantization.md like `See the available quantization schemes [here](https://github.com/pytorch/torchchat/blob/main/torchchat/utils/quantize.py#L1260-L1266).`. Could you please help update it to show the quantization schemes examples?

### Versions

#

yanbing-j:
cc @mikekgfb @mingfeima

yanbing-j:
In main branch, torchchat/utils/quantize.py only has 942 lines.

Jack-Khuu:
Ah good catch

PR #1392, opened 2024-11-22
Update quantization.md link to quantize.py

Jack-Khuu:
https://github.com/pytorch/torchchat/issues/1385


docs/quantization.md
vmpuri, line 1: Should this be in `/blob/main`?

Jack-Khuu: not sure on the question, blob is GH's identifier for type of UI

vmpuri (approved):

PR #1398, opened 2024-12-04
Bump stale PT pin to 20241113

Jack-Khuu:
The current pin  of 20241002 is no longer available as a prebuilt pt wheel. This bumps it to 20241020, while we work on a fresher bump in #1367


Jack-Khuu:
Closing infavor of https://github.com/pytorch/torchchat/pull/1367

PR #1400, opened 2024-12-06
Bump torch pin to 20241010

larryliu0820:
As titled


Jack-Khuu:
You'll need this change to get around the ggreganov BC break: https://github.com/pytorch/torchchat/blob/f38d3f4bb675694de4297d6994ee88743e7feb32/.github/workflows/pull.yml#L734

Jack-Khuu:
Also ET pin needs to be bumped: https://github.com/pytorch/torchchat/blob/6e8bfb181a4afe11a919f29aac98eb3ae94ef424/install/.pins/et-pin.txt

Jack-Khuu (approved):
Thanks for the patch

Jack-Khuu:
See https://github.com/pytorch/torchchat/actions/runs/12204230280/job/34048909674?pr=1367 for a more recent bump

PR #1415, opened 2024-12-10
Add torchao mps ops

manuelcandales:
This PR adds the quantization scheme linear:afpwx. It quantizes only the weights in a groupwise manner with a specified bitwidth and groupsize. It takes arguments bitwidth (1, 2, 3, 4, 5, 6, 7) and groupsize (32, 64, 128, 256).

To use linear:afpwx, you must first set up the torchao mps experimental kernels. These will only work on a device with Apple Silicon.

From the torchchat root directory, run
```
bash torchchat/utils/scripts/build_torchao_ops.sh mps
```

Notice that this quantization scheme is currently implemented only for device mps.
```
python3 torchchat.py generate stories110M --device mps --dtype float32 --quantize '{"linear:afpwx": {"bitwidth": 4, "groupsize": 256}}' --prompt "Once upon a time,"
```


docs/quantization.md
```diff
@@ -196,6 +196,31 @@ Note: only the ExecuTorch C++ runner in torchchat when built using the instructi
...
 ```
 
+## Experimental TorchAO MPS lowbit kernels
+
+WARNING: These kernels only work on devices with Apple Silicon.
+
+### Use
+
+#### linear:fpaxw
+The quantization scheme linear:fpaxw quantizes only the weights in a groupwise manner with a specified bitwidth and groupsize.
```
metascroy, line 206: Should we keep the naming convention that torchchat uses (of "a" followed by type and "w" followed by type).  This started with a8w4dq before I added any kernels.  In your case this would be something like afpwx?

torchchat/utils/scripts/install_utils.sh
metascroy, line 1: Can you add a CI test for the MPS kernels to make sure they install and run?

See https://github.com/pytorch/torchchat/blob/main/.github/workflows/pull.yml#L1060 as an example.

docs/quantization.md
```diff
@@ -196,6 +196,31 @@ Note: only the ExecuTorch C++ runner in torchchat when built using the instructi
...
+From the torchchat root directory, run
+```
+sh torchchat/utils/scripts/build_torchao_ops.sh mps
+```
+
+### Examples
+
+#### Eager mode
+```
+python3 torchchat.py generate stories110M --device mps --dtype float32 --quantize '{"linear:fpaxw": {"bitwidth": 4, "groupsize": 256}}' --prompt "Once upon a time," --num-samples 5
```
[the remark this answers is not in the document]
manuelcandales: The metal lowbit kernels run with ExecuTorch as well (llama runner can use them). However, my aim in this torchchat PR was only to enable eager. I plan to have a follow up PR to enable them via the torchchat ET path as well. But I prefer to keep it modular.

docs/quantization.md
manuelcandales: Ok, this makes sense.

docs/quantization.md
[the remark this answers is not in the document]
manuelcandales: I added a sentence in the setup section, clarifying that currently torchchat can only use them on Eager mode

torchchat/utils/scripts/install_utils.sh
manuelcandales: done!

metascroy (approved):

Issue #1416, opened 2024-12-11
MacOS test falsely uses MPS, fails and is misreported as passing


mikekgfb:
### 🐛 Describe the bug

#1415, #1404 and all other PRs fail on test-readme-macos when torchchat apparently falsely tries to load the model to MPS.  
Due to #1315 , the test don't report as failed, so things get committed anyway.

I don't know why test-readme-macos would try to load into MPS.  There's a multi-layered story here, where virtualized Mac does not support MPS (I think because most likely there's no MMU for MPS, so you can't virtualize MPS).  MPS is however still reported as available by the OS, and hence a simple check `torch.backends.mps.is_available():` is not sufficient because pytorch thinks that MPS is actually available (but any and all memory allocations should fail).

We're trying to fix this by doing an allocation of a tensor in MPS memory and see if that succeeds or fails in `is_mps_available()` here => https://github.com/pytorch/torchchat/blob/main/torchchat/utils/build_utils.py#L269 when `get_device_str()` is looking to check if MPS is available, and the fastest device should be returned as MPS.

Ideally, this should be fixed in the expansion `get_device_str()` and `is_mps_available()` functions, together with #1315.

### Versions

problem occurs in github ci/cd

Issue #1315, opened 2024-10-18
False Positive on CI tests: test-readme


Jack-Khuu:
### 🐛 Describe the bug

In CI, there are a few tests that should be flagged as failing, but are currently marked as green. 

Specifically they seem to revolve around the test-readme unittests: see surfacing PR for examples (https://github.com/pytorch/torchchat/pull/1309)

* https://github.com/pytorch/torchchat/actions/runs/11375390717/job/31652547554?pr=1309
* https://github.com/pytorch/torchchat/actions/runs/11375390708/job/31652543330?pr=1309

### Versions

NA

mikekgfb:
@seemethere @malfet can you please have a look as to why these tests are not flagged as failing? 

ditto, from yesterday - https://github.com/pytorch/torchchat/actions/runs/11634438558/job/32413340991?pr=1339 is shown as passed even though commands in the test failed and aborted with an error indication?

mikekgfb:
And still happening.

One possible explanation might be that somewhere the code is catching the exception, pretty printing it, and then exit with a non-error code because the fails such as https://github.com/pytorch/torchchat/actions/runs/12243820522/job/34154220414?pr=1404 show that the code continues executing (see test run dump below) despite the fact that the first command in the generated test is `  set -eou pipefail` which instructs the shell to immediately abort on  the first error, and report a failure to the caller.  (And recursively, the workflows also use this setting to cascade the first test error all the way to the top and make a test fail)

```
+ python3 torchchat.py generate stories15M --prompt 'write me a story about a boy and his bear'
## Running via PyTorch 
  Downloading https://huggingface.co/karpathy/tinyllamas/resolve/main/stories15M.pt...
  Downloading https://github.com/karpathy/llama2.c/raw/master/tokenizer.model...
  NumExpr defaulting to 6 threads.
  PyTorch version 2.6.0.dev20241013 available.
  Moving model to /Users/runner/.torchchat/model-cache/stories15M.
  
  Downloading builder script:   0%|          | 0.00/5.67k [00:00<?, ?B/s]
  Downloading builder script: 100%|██████████| 5.67k/5.67k [00:00<00:00, 5.30MB/s]
  Traceback (most recent call last):
    File "/Users/runner/work/torchchat/torchchat/torchchat.py", line 96, in <module>
  Using device=mps 
  Loading model...
      generate_main(args)
    File "/Users/runner/work/torchchat/torchchat/torchchat/generate.py", line 1235, in main
[...]

PR #1417, opened 2024-12-11
bandaid for run-readme-pr-macos.yml incorrectly loading to MPS

mikekgfb:
as per #1416 torchchat on hosts without MPS (which is all github hosts which use kvm to virtualize MacOS, but not MPS) should choose CPU as "fast" device.  The logic is present (see discussion in #1416 ), but either not fully functional (that would be the easier one to fix, just print the result of get_device_str and fix the code!) or specifically ignored on load in torch/serialization.py (If this is the case, we're effectively looking at a core PyTorch issue....)

In the meantime, this bandaid just forces the use of CPU on MacOS tests, to make MacOS tests run on CPU -- labeit hsortcircuiting test/execution of the "fast" device logic.  Not ideal, but some testing beats no testing.


mikekgfb:
PS: This fix should make #1315 not trigger in this case (because tests should succeed, so we don't run into quiet fail), but doesn't do anything to actually address the #1315 issue.

Jack-Khuu (approved):

PR #1432, opened 2024-12-19
Granite 3.0 / 3.1 dense support

gabe-l-hart:
## Description

This PR adds support for the dense Granite models in the 3.0 and 3.1 collections

## Changes

* Add granite architecture parameters
* Use granite multipliers where appropriate only if set in config
* Add `model_params` for each of the new granite models
* Add entries in `model.json` for each of the new granite models


Jack-Khuu:
Just initial glance so far, but wow this is really clean

Jack-Khuu (approved):

Issue #1429, opened 2024-12-19
[KNOWN BUG] AOTI Inference via cpp Runner bug


Jack-Khuu:
### 🐛 Describe the bug

As of https://github.com/pytorch/torchchat/pull/1426, `torchchat/main` is failing 2 jobs related to AOTI inference via c++ runners. 

### Versions

https://github.com/pytorch/torchchat/commit/fd1857a6deb9c425527d4fd69fba2c5ca5641f93

mikekgfb:
Seems to be a problem in sentencepiece https://github.com/pytorch/torchchat/blob/cc0ffce74f50fd9d47db495aafa50d498dbf447d/tokenizer/sentencepiece.cpp#L39-L43

PR #1435, opened 2024-12-20
Wording sentencepiece.cpp

mikekgfb:
. before newline.  Reformat file name to make clear . is not part of filename

### DIFF
--- a/tokenizer/sentencepiece.cpp
+++ b/tokenizer/sentencepiece.cpp
@@ -38,7 +38,7 @@ void SPTokenizer::load(const std::string& tokenizer_path) {
   // read in the file
   const auto status = _processor->Load(tokenizer_path);
   if (!status.ok()) {
-    fprintf(stderr, "couldn't load %s\n. If this tokenizer artifact is for llama3, please pass `-l 3`.", tokenizer_path.c_str());
+    fprintf(stderr, "Could not load `%s`.\n If this tokenizer artifact is for llama3, please pass `-l 3`.", tokenizer_path.c_str());
     exit(EXIT_FAILURE);
   }
   // load vocab_size, bos_tok, eos_tok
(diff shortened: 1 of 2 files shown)


Jack-Khuu (approved):

tokenizer/sentencepiece.cpp
```diff
@@ -38,7 +40,13 @@ void SPTokenizer::load(const std::string& tokenizer_path) {
   // read in the file
   const auto status = _processor->Load(tokenizer_path);
   if (!status.ok()) {
-    fprintf(stderr, "couldn't load %s\n. If this tokenizer artifact is for llama3, please pass `-l 3`.", tokenizer_path.c_str());
+    // Execute 'ls -al' on the tokenizer path
```
Jack-Khuu, line 43: Thanks for adding the print, great for debugging

Looks like the `ls` is spitting out the root torchchat directory instead of tokenize path which is curious

mikekgfb: That would explain why the tokenizer can't be loaded, and the AOTI tests keep failing. #1429

mikekgfb: I added set -x to the command to echo the ls to make absolutely sure the path is not getting corrupted somehow (not sure how it would, but belts and suspenders)

Jack-Khuu: Neutral signal- looks like the arg is not being picked up by ls (which would explain why it just shows PWD)

mikekgfb: That is sooooo weird! Want to add a print of command and rerun? Maybe there’s some magic character that causes indigestion for the shell running La; and the tokenizer model load?

mikekgfb: Added print of command before execution

mikekgfb: Split C style strong conversion and c++ const as convert followed by append

Jack-Khuu:
Thanks again for helping with the debug. 

@larryliu0820 was able to get his tokenizer changes back in so we can either rebase or close this one
https://github.com/pytorch/torchchat/pull/1443

mikekgfb:
> Thanks again for helping with the debug.
> 
> @larryliu0820 was able to get his tokenizer changes back in so we can either rebase or close this one #1443

Awesome. I suggest we close this one.

Issue #1436, opened 2024-12-22
If scripts need `bash`, don't say to use `sh`


swolchok:
### 🐛 Describe the bug

On Debian systems, sh isn't bash, it's [dash](https://en.wikipedia.org/wiki/Almquist_shell#Dash). I haven't tested every script, but https://github.com/pytorch/torchchat/blob/main/docs/quantization.md says to run `sh torchchat/utils/scripts/build_torchao_ops.sh`, but this script fails unless run with bash on my Raspberry Pi 5.

### Versions

Collecting environment information...
PyTorch version: 2.6.0.dev20241218+cpu
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: Debian GNU/Linux 12 (bookworm) (aarch64)
GCC version: (Debian 12.2.0-14) 12.2.0
Clang version: Could not collect
CMake version: version 3.31.2
Libc version: glibc-2.36

Python version: 3.11.2 (main, Sep 14 2024, 03:00:30) [GCC 12.2.0] (64-bit runtime)
Python platform: Linux-6.6.51+rpt-rpi-2712-aarch64-with-glibc2.36
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Architecture:                         aarch64
CPU op-mode(s):                       32-bit, 64-bit
Byte Order:                           Little Endian
CPU(s):                               4
On-line CPU(s) list:                  0-3
Vendor ID:                            ARM
Model name:                           Cortex-A76
Model:                                1
Thread(s) per core:                   1
Core(s) per cluster:                  4
Socket(s):                            -
Cluster(s):                           1
Stepping:                             r4p1
CPU(s) scaling MHz:                   100%
CPU max MHz:                          2400.0000
CPU min MHz:                          1500.0000
BogoMIPS:                             108.00
Flags:                                fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp
L1d cache:                            256 KiB (4 instances)
L1i cache:                            256 KiB (4 instances)
L2 cache:                             2 MiB (4 instances)
L3 cache:                             2 MiB (1 instance)
[...]

PR #1437, opened 2024-12-22
Update sh -> bash in quantization.md

mikekgfb:
Resolve one instance of #1436 where we say sh but mean bash (sh is not bash on every system).


Jack-Khuu (approved):
Thanks for the fast fix

PR #1438, opened 2024-12-22
Create distributed.md

mikekgfb:
Initial documentation for use of distributed inference w/ torchchat. @mreso please review and update as appropriate.

### DIFF
--- a/docs/distributed.md
+++ b/docs/distributed.md
@@ -1,0 +1,113 @@
+# Distributed Inference with torchchat
+
+torchchat suports distributed inference for large language models (LLMs) on GPUs seamlessly. 
+At present, torchchat supports distributed inference using Python only.
+
+## Installation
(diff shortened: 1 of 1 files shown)


kwen2501 (approved):
Looks great! Thanks so much for writing it up!

mreso (approved):
Thanks @mikekgfb looks good, the response of the server is a bit different but thats true for the non distributed case too.

mikekgfb:
@mreso @kwen2501 @lessw2020 do you have a perspective what list of models will run with distributed? I know we had a separate model doc

Any attempts at 405B? Wer should def add some of the super large models to the models.Json so they're accessible for distributed inference?

Also, should we create a spare table here, or we could put a second column in the main readme for "distributed inference"? Next to "mobile" column?

PR #1439, opened 2024-12-23
Update run-docs to avoid code duplication

mikekgfb:
Update run-docs to avoid duplicate code (hygiene for scalability with more doc files)

Also, add support for command line options without space for runner (in line with how POSIX treats commandline arguments) , to offer a more portable way to specify rewrite from readme.md to tests with updown.py.


mikekgfb:
@mreso  @kwen2501 @lessw2020 do you have a perspective what list of models will run with distributed? I know we had a separate model doc

Any attempts at 405B? Wer should def add some of the super large models to the models.Json so they're accessible for distributed inference?

Also, should we create a spare table here, or we could put a second column in the main readme for "distributed inference"? Next to "mobile" column?

.ci/scripts/run-docs
```diff
@@ -1,145 +1,67 @@
...
-        echo "::group::Run distributed"
-        echo "*******************************************"
-        cat ./run-distributed.sh
-        echo "*******************************************"
-        bash -x ./run-distributed.sh
-        echo "::endgroup::"
-fi
+# Pre-initialize variables
+filepath=""
+parameters="--replace 'llama3:stories15M,-l 3:-l 2' --suppress huggingface-cli,HF_TOKEN"
```
Jack-Khuu, line 11: Looks like we have quotation bug making the tests fail silently

```
 python3 torchchat/utils/scripts/updown.py --file torchchat/utils/docs/evaluation.md --replace ''\''llama3:stories15M,-l' 3:-l '2'\''' --suppress huggingface-cli,HF_TOKEN
  usage: updown.py [-h] [-f FILENAME] [-p PREDICATE] [-r REPLACE] [-s SUPPRESS]
                   [-e] [-g]
  updown.py: error: unrecognized arguments: 3:-l 2'
```

mikekgfb: On what system is that?  There's definitely something weird going on here.  Angela committed a change that we don't need -l 2/3 anymore, so we would not have to switch around that flag?  the problem originates because of the space in the arg that should be properly escaped, but some shell does not seem to handle this gracefully.  

And yeah, all errors are being suppressed on run-docs commands, so it's been attracting a ton of fails that are surfacing when we look closely

Jack-Khuu: https://github.com/pytorch/torchchat/actions/runs/12996075947/job/36258706324?pr=1439
https://github.com/pytorch/torchchat/actions/runs/12996075978/job/36258715248?pr=1439

> Angela committed a change that we don't need -l 2/3 anymore, so we would not have to switch around that flag?

https://github.com/pytorch/torchchat/pull/1159 Correct for AOTI we should't need the flag

Jack-Khuu: > yeah, all errors are being suppressed on run-docs commands, so it's been attracting a ton of fails that are surfacing when we look closely

Surfacing suppressed errors is a good thing :)

mikekgfb: > https://github.com/pytorch/torchchat/actions/runs/12996075947/job/36258706324?pr=1439 https://github.com/pytorch/torchchat/actions/runs/12996075978/job/36258715248?pr=1439
> 
> > Angela committed a change that we don't need -l 2/3 anymore, so we would not have to switch around that flag?
> 
> #1159 Correct for AOTI we should't need the flag

et_run still needs the flag?  Can we do something for et_run too?  It would both help the users and simplify the cleanup ;)

PS: would -l3 work based on the argline parser?  Coz that one doesn't trigger the space in arguments issue that leads to the problem we're seeing.  (I think it's a badly implemented shell, but it is what it is, and no point to visit "is there a better shell" and then investigate how to rebuild git around it.... Pragmatically, cheapest way we get over that hump, since iit's sorta a nuisance to say -l anyway.... so if we can just kill it off, that's the best path!)

mikekgfb: So, I've removed -l 3 from aoti_run, and rewritten it as -l3 for et_run in the hope that the parser is rational and can handle traditional POSIX command line processing (where this space is historically optional).  If we don't need -l 3 on et_run either, that would be best for our users anyway.

@Jack-Khuu

Jack-Khuu:
Thanks for this, kicked off the CI

Jack-Khuu (approved):

.ci/scripts/run-docs
Jack-Khuu: Thanks for the workaround, agree we should add it for ET (pretty straight forward)

https://github.com/pytorch/torchchat/issues/1484

Issue #1413, opened 2024-12-10
llama3.2-1b-base gives 'RuntimeError: no tokenizer was found'


drtonyr:
### 🐛 Describe the bug

I'm following the basic 5+1 line instructions to get torchchat going with llama3.2-1b-base and it fails with `RuntimeError: no tokenizer was found at /home/tonyr/.torchchat/model-cache/meta-llama/Meta-Llama-3.2-1B/tokenizer.model`.  The 6 lines I'm using are from the README.md:
```text
git clone https://github.com/pytorch/torchchat.git
cd torchchat
python3 -m venv .venv
source .venv/bin/activate
./install/install_requirements.sh
python3 torchchat.py generate llama3.2-1b-base --prompt "write me a story about a boy and his bear"
```
I also have llama3.2-3b-base which exhibits the same behaviour.  Full logs are below, note I already have llama3.2-1b-base downloaded:
```text
(.venv) think1 tonyr: ls -lta /home/tonyr/.torchchat/model-cache/meta-llama/Meta-Llama-3.2-1B/
total 2424908
drwxrwxr-x 4 tonyr tonyr       4096 Dec  8 20:45 ..
drwxrwxr-x 4 tonyr tonyr       4096 Dec  8 20:40 .
drwxrwxr-x 2 tonyr tonyr       4096 Dec  8 20:40 original
-rw-rw-r-- 1 tonyr tonyr 2471677246 Dec  8 20:40 model.pth
-rw-rw-r-- 1 tonyr tonyr    9085657 Dec  8 20:38 tokenizer.json
-rw-rw-r-- 1 tonyr tonyr    2183982 Dec  8 20:38 tokenizer.model
-rw-rw-r-- 1 tonyr tonyr       1519 Dec  8 20:38 .gitattributes
-rw-rw-r-- 1 tonyr tonyr      50500 Dec  8 20:38 tokenizer_config.json
-rw-rw-r-- 1 tonyr tonyr        301 Dec  8 20:38 special_tokens_map.json
-rw-rw-r-- 1 tonyr tonyr        185 Dec  8 20:38 generation_config.json
-rw-rw-r-- 1 tonyr tonyr       7712 Dec  8 20:38 LICENSE.txt
-rw-rw-r-- 1 tonyr tonyr        843 Dec  8 20:38 config.json
-rw-rw-r-- 1 tonyr tonyr      41239 Dec  8 20:38 README.md
[...]

PR #1401, opened 2024-12-06
Use pytorch-labs/tokenizers and remove tokenizer/

larryliu0820:
Summary: Use our shiny new repo
https://github.com/pytorch-labs/tokenizers.

Test Plan: Rely on CI jobs

Reviewers:

Subscribers:

Tasks:

Tags:

Issue #1440, opened 2024-12-23
Migrate Tokenizer Components to utilize pytorch-labs/tokenizers


Jack-Khuu:
### 🚀 The feature, motivation and pitch

@larryliu0820 has created a new shared repository for [hosting tokenizer definitions](https://github.com/pytorch-labs/tokenizers). 

The initial migration attempt was reverted in  https://github.com/pytorch/torchchat/pull/1414 due to a tokenizer issue flagged in https://github.com/pytorch/torchchat/issues/1413, but should be straightforward to debug and reland

**Task**: Taking inspiration from https://github.com/pytorch/torchchat/pull/1401, reattempt this migration

PR #1443, opened 2024-12-26
[retry] Use pytorch-labs/tokenizers and remove tokenizer/ (#1401)

larryliu0820:
Summary: Retry of #1401

Test Plan: Re-run the repro command in #1413:

```
python3 torchchat.py generate llama3.2-1b-base --prompt "write me a story about a boy and his bear"

```

Reviewers:

Subscribers:

Tasks:

Tags:


torchchat/utils/scripts/build_native.sh
```diff
@@ -93,7 +93,7 @@ popd
 if [[ "$TARGET" == "et" ]]; then
     cmake -S . -B ./cmake-out -DCMAKE_PREFIX_PATH=`python3 -c 'import torch;print(torch.utils.cmake_prefix_path)'` -DLINK_TORCHAO_OPS="${LINK_TORCHAO_OPS}" -DET_USE_ADAPTIVE_THREADS=ON -DCMAKE_CXX_FLAGS="-D_GLIBCXX_USE_CXX11_ABI=1" -G Ninja
 else
-    cmake -S . -B ./cmake-out -DCMAKE_PREFIX_PATH=`python3 -c 'import torch;print(torch.utils.cmake_prefix_path)'` -DLINK_TORCHAO_OPS="${LINK_TORCHAO_OPS}" -DCMAKE_CXX_FLAGS="-D_GLIBCXX_USE_CXX11_ABI=0" -G Ninja
+    cmake -S . -B ./cmake-out -DCMAKE_PREFIX_PATH=`python3 -c 'import torch;print(torch.utils.cmake_prefix_path)'` -DLINK_TORCHAO_OPS="${LINK_TORCHAO_OPS}" -DCMAKE_CXX_FLAGS="-D_GLIBCXX_USE_CXX11_ABI=1" -G Ninja
```
larryliu0820, line 96: Getting "model file path should not be empty." from sentencepiece processor.

According to chatgpt:

The `_GLIBCXX_USE_CXX11_ABI` macro is related to the Application Binary Interface (ABI) used by the GNU C++ Standard Library.
In 2011, the C++11 standard introduced a new ABI for C++ objects, which changed the way that standard library objects, including `std::string`, are laid out in memory. This new ABI is enabled by default in GCC 5 and later versions.
However, when `_GLIBCXX_USE_CXX11_ABI=0` is defined, it tells the compiler to use the older, pre-C++11 ABI instead. This can cause issues with `std::string` objects because the older ABI uses a different memory layout for these objects.
When you pass a `const char*` pointer to a function that expects a `std::string` object, the compiler needs to create a temporary `std::string` object from the pointer. However, if the ABI is set to the older version, the compiler may not be able to correctly create this temporary object, resulting in an empty string being passed to the function.
This issue is likely due to the fact that the older ABI uses a different memory layout for `std::string` objects, which includes a pointer to the string data, a length field, and a capacity field. When the compiler creates a temporary `std::string` object from a `const char*` pointer, it needs to initialize these fields correctly. However, if the ABI is set to the older version, the compiler may not be able to correctly initialize these fields, resulting in an empty string.

Jack-Khuu (approved):

Jack-Khuu:
Thanks for the fix!!

xrf: https://github.com/pytorch/torchchat/issues/1440

PR #1445, opened 2024-12-27
Update README.md to include granite

mikekgfb:
@gabe-l-hart @Jack-Khuu adding to table since #1262 is complete.  Does torchchat need anything else before it's appropriate to say granite is supported?


gabe-l-hart (approved):
Thank you! These Granite models are good to go now

PR #1447, opened 2024-12-30
Change role to user for user prompts

vladoovtcharov:
User prompts are currently being set to the "system" role instead of "user"


Jack-Khuu:
Looks like this one slipped through the cracks; acking will test and merge!!

Jack-Khuu (approved):
Thanks for the fix!

PR #1450, opened 2025-01-01
Update License Copyright Year

JasonnnW3000:
Make a new PR to the copyright as suggested, thank you!

Yearly copyright bump. Padding my changed LOC! 😛

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2024 Meta
+Copyright (c) 2024-2025, Meta Platforms, Inc.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:


justinfagnani:
This PR is part of a PR spam campaign. I would close it.

xurizaemon:
There is no reason for introducing an end date to the copyright statement - my understanding is that it's sufficient to state the date of first copyright, but it's a common practice to include the "until `date('Y')` part, which creates an additional maintenance task of updating the year annually.

Nobody seriously believes things drop out of copyright if this busywork task gets missed, and I think we'd all be better off just using the initial year and skipping the extra task.

I'm not a user of this project - take my input with a large grain of salt, I have no legal qualifications in your jurisdiction. I'm just here because I'm one of the ~1.5K maintainers who got a "just padding my commit graph" PR from @JasonnnW3000, but I think we'd all be better off with one less annual task in 2026 and beyond 😉

PR #1455, opened 2025-01-10
Update aoti calls to utilize new export and packaging APIs

angelayi:


torchchat/export.py
```diff
@@ -72,16 +72,17 @@ def export_for_server(
...
             example_inputs,
             dynamic_shapes=dynamic_shapes,
-            options=options,
         )
+        options=options,
 
         if package:
-            from torch._inductor.package import package_aoti
-            path = package_aoti(output_path, path)
+            path = torch._inductor.aoti_copmile_and_package(ep, package_path=output_path, inductor_configs=options)
```
mikekgfb, line 83: typo: aoti_copmile_and_package -> aoti_compile_and_package

Jack-Khuu:
Thanks for the update, mind adding a bit of context to the description?

Looks like the error is legit? An arg mismatch

PR #1457, opened 2025-01-14
Change role in chat to user

nlpfollower:
Shouldn't this be a "user" role instead?


nlpfollower:
Closing, duplicate of #1447.

PR #1458, opened 2025-01-15
Update ARM CPU experimental kernels from AO to leverage pip install

metascroy:
* torchao experimental CPU kernels are now installed and loaded automatically by pip.
* Switch quantization to use new quantize_ API


docs/quantization.md
```diff
@@ -138,7 +140,9 @@ The quantization scheme embedding:wx quantizes embeddings in a groupwise manner
 You should expect high performance on ARM CPU if groupsize is divisible by 32.  With other platforms and argument choices, a slow fallback kernel will be used.  You will see warnings about this during quantization.
 
 ### Setup
-To use linear:a8wxdq and embedding:wx, you must set up the torchao experimental kernels.  These will only work on devices with ARM CPUs, for example on Mac computers with Apple Silicon.
+If you are using the torchao ops from python, they are available out of the box on a Mac with Apple Silicon, and you can skip these setup steps.
```
Jack-Khuu, line 143:
```suggestion
If you are using the torchao ops from python (i.e not with a C++ runner), they are available out of the box on a Mac with Apple Silicon, and you can skip these setup steps.
```

torchchat/utils/quantize.py
```diff
@@ -117,7 +129,47 @@ def quantize_model(
...
+                if get_precision() != torch.float32:
+                    print(f"Quantizer {quantizer} requires float32 inputs, but received {get_precision()}.  Changing dtype to float32.  Note that after quantization, the weights will be lowbit integers, not float32.")
+                    set_precision(torch.float32)
+
+                group_size = q_kwargs["groupsize"]
+                bit_width = q_kwargs["bitwidth"]
+                has_weight_zeros = q_kwargs["has_weight_zeros"]
+                granularity = PerRow()
+                if group_size != -1:
+                    granularity = PerGroup(group_size)
```
Jack-Khuu, line 142:
```suggestion
                granularity = PerRow() if group_size == -1 else PerGroup(group_size) 
```

nit

torchchat/utils/quantize.py
```diff
@@ -117,7 +129,47 @@ def quantize_model(
...
+                        int8_dynamic_activation_intx_weight(
+                            weight_dtype=weight_dtype,
+                            granularity=granularity,
+                            has_weight_zeros=has_weight_zeros,
+                            layout=PackedLinearInt8DynamicActivationIntxWeightLayout(),
+                        ),
+                    )
+                except Exception as e:
+                    print("Encountered error during quantization: {e}")
+                    print("Trying with PlainLayout")
```
Jack-Khuu, line 157: Is there a specific case we are anticipating here?

torchchat/utils/quantize.py
```diff
@@ -916,26 +970,11 @@ def quantized_model(self) -> nn.Module:
         torchao_experimental_quant_api
     )
     from torchao_experimental_quant_api import (
-        Int8DynActIntxWeightLinearQuantizer,
-        IntxWeightEmbeddingQuantizer,
         UIntxWeightOnlyLinearQuantizer,
     )
```
Jack-Khuu, line 974:
```suggestion
        from torchao_experimental_quant_api import UIntxWeightOnlyLinearQuantizer
```

Drops down to single import, can lose the (...)

Jack-Khuu (approved):
Thanks for making the pip install work with the subclass APIs!!

Jack-Khuu:
cc: @manuelcandales 

Can have please for MPS? 🥺🥺 (Separate PR)

torchchat/utils/quantize.py
metascroy: If using this on cuda machine of x86 machine

Jack-Khuu:
Awaiting https://github.com/pytorch/executorch/pull/7759?fbclid=IwZXh0bgNhZW0CMTEAAR1561BK21eYspDfjCt0Zgx7MUetM8yl4HS_NaL-1VEyHw1WLwoWxJ_nom0_aem_5DdCJGYQx6zSAcEOn8Vfwg

metascroy:
@Jack-Khuu did you update the version AO uses in ET?

Jack-Khuu:
Yup https://github.com/pytorch/executorch/commit/9836b39fe690e1906f133b4a233863149c30d499 Points to https://github.com/pytorch/ao/commit/11333ba2cb5c4e792bc4f5c0d70c12991f972008

Jack-Khuu:
Hi @nikhil-arm, we're still planning to land this

Can you share the specific commit hashes y'all need?

Jack-Khuu:
@nikhil-arm We've bumped the AO pin on main. 
Please let me know if you there's any additional support needed to unblock KleidiAI kernels

install/install_requirements.sh
```diff
@@ -117,9 +117,11 @@ fi
 
 # For torchao need to install from github since nightly build doesn't have macos build.
 # TODO: Remove this and install nightly build, once it supports macos
+# USE_CPP=1 indicates that the torchao experimental aten kernels will be built and loaded
+# if on Mac with Apple Silicon
 (
   set -x
-  $PIP_EXECUTABLE install git+https://github.com/pytorch/ao.git@7d8794622f3ac7ffa98761314019a20fba06edef
+  USE_CPP=1 $PIP_EXECUTABLE install git+https://github.com/pytorch/ao.git@7d8794622f3ac7ffa98761314019a20fba06edef
```
metascroy, line 124: Shouldn't this match what's in install/.pins/torchao-pin.txt?

Jack-Khuu:
After a suite of rebases, pinbumps, and splitting up tests up we know what we're tackling:
- `test-torchao-experimental-cpp (macos-14-xlarge)`: Tests the AOTI runner and likely failing (also in main) due to not linking to the LibOMP from torch as @malfet mentioned in https://github.com/pytorch/torchchat/pull/1493.

- `test-torchao-experimental-et (macos-14-xlarge)`: Tests the ET runner; looks like a install bug where USE_CPP isn't set, but will likely run into the same LibOMP issue as above

metascroy:
> Hello @metascroy @Jack-Khuu , what is the plan to get this in mainline? We would like to use KleidiAI kernels from aten via this quantizer path. Let us know if we need to raise a new PR ?

Sorry about the delay @nikhil-arm.

@Jack-Khuu let's try to get this landed within the next week.  Bumping the ao pin in torchchat had various conflicts with the CI, but I think we can dedicate to making this work.

I think it does make sense to first land https://github.com/pytorch/ao/pull/1836 in torchao before bumping because they've already deprecated the old quantizers in quantize_.

Jack-Khuu (approved):
Re-review and things are looking great

Thanks again

Jack-Khuu:
Hi @nikhil-arm The experimental kernels from AO are now in. Can you share a link to the KleidiAI Kernels (either here or a GH issue if you can share more context)?

metascroy:
> Hi @nikhil-arm The experimental kernels from AO are now in. Can you share a link to the KleidiAI Kernels (either here or a GH issue if you can share more context)?

For context for both @nikhil-arm and @Jack-Khuu.  There are two locations of KleidiAI kernels in PyTorch/torchao now.

PyTorch has a 4-bit quantized linear op backed by KleidiAI.  I think @nikhil-arm wants to enable this in torchchat.  To do this, you need to pass "aten" to the target in PackedLinearInt8DynamicActivationIntxWeightLayout(target="aten") here: https://github.com/pytorch/torchchat/blob/main/torchchat/utils/quantize.py#L150.  I will leave it to @nikhil-arm to put up a PR to pipe this through.  This should then work with most surfaces, except ExecuTorch, although I don't know if it's been tested in anything other than eager mode.

We have also pulled in KleidiAI kernels in torchao itself.  These should work on all surfaces, including ExecuTorch.  To enable these, we need to add the flag TORCHAO_BUILD_KLEIDIAI=1 before the pip install: https://github.com/pytorch/torchchat/blob/8a9a644a1ff83595fd12ab07930ae44a418ceb73/install/install_torchao.sh#L38 

When enabled, PackedLinearInt8DynamicActivationIntxWeightLayout() will use the KleidiAI kernel when supported (4-quantization, has_weight_zeros=false), and use our native torchao kernels in other cases.  You can see what kernel is being selected at runtime by setting the environment variable TORCH_CPP_LOG_LEVEL=Info.  Eventually we want to enable this flag by default, but I have it disabled right now for two reasons: 1) the flag increases the install time because it clones/builds KleidiAI; 2) there is limited perf benefit to enabling it now based on the KleidiAI kernels we have enabled in torchao (GEMV neondot kernels) vs. the native torchao Arm kernels (also GEMV neondot kernels).

Issue #1032, opened 2024-08-15
RFC: Make quantization a first class feature


byjlw:
### 🚀 The feature, motivation and pitch

torchchat provides quantization functionality but the interaction isn't ideal. 
### Currently
Currently you can run generate with a --quantization flag and it quantize the model before running inference. This means that you have to quantize (expensive) every time you want to run generate. It also means you can't ask for a quantized model for use in another project. 
```
python3 torchchat.py generate llama2 --quantize '{"linear:int4": {"groupsize": 128}, "precision": {"dtype":"float16"}, "executor":{"accelerator":"mps"}}' --prompt "Once upon a time," --max-new-tokens 256 --num-samples 3 --seed 42
```

### Expectation
Ideally you should be able to save/cache the quantized model during generate *and* run a command to just quantize and get the outputted model

**Add a quantize command**
```
python3 torchchat.py quantize llama2  '{"linear:int4": {"groupsize": 128}, "precision": {"dtype":"float16"}, "executor":{"accelerator":"mps"}}' --output-file llama2-4b.pt
``` 
- the quantize.md file should enumerate all options available in a table along with a suggested option
- the quantize.md file should provide examples
- the readme.md file should include one example for 4bit (often used for benchmarking)
- optional output-file (if flag isn't present just store it in our default model location with a well defined scheme `{modelname}-{quant}.pt`)

**Cache the quantized versions when you run using --quantize**
- the `list` command should show the quantized versions
- the `remove` command should delete the quantized version

### Additionally
- common quantization recipes are stored in a json file and can be passed in
 - `torchchat/config/quantization/llama-4bit.json`
 - with the contents `{"linear:int4": {"groupsize": 128}, "precision": {"dtype":"float16"}, "executor":{"accelerator":"mps"}}`
 - allowing for the command `python3 torchchat.py quantize llama2 --config config/quantization/llama-4bit.json --output-file llama2-4bit.json`

PR #1465, opened 2025-01-18
Add export --output-snapshot snap.tc, and --snapshot snap.tc

mikekgfb:
Add ability to save and restore quantized models #1032 

```
mgschwind@mgschwind-mlt torchchat % python3 torchchat.py generate stories15M --quant torchchat/quant_config/desktop.json --prompt "once upon a time"
NumExpr defaulting to 12 threads.
PyTorch version 2.6.0.dev20241218 available.
Unabled to import torchao experimental quant_api with error:  [Errno 2] No such file or directory: '/Users/mgschwind/tc/torchchat/torchao-build/src/ao/torchao/experimental/quant_api.py'
Using device=mps 
Loading model...
Time to load model: 0.18 seconds
Quantizing the model with: {'executor': {'accelerator': 'fast'}, 'precision': {'dtype': 'fast16'}}
Time to quantize model: 0.00 seconds
once upon a time, there was a little girl named Lily. She loved to play outside in the park with her friends. One day, Lily saw a big, scary dog. She was frightened and didn't know what to do. 
Lily's friend, Timmy, saw her and asked, "What's wrong, Lily?" 
"I'm frightened of the dog," Lily said. 
Timmy said, "Don't worry, I'll call my mom and she will come to save us." 
After Timmy called, he ran to Lily's mom and told her about the scary dog. Lily's mom called the street workers to come and take the dog away. 
Lily was happy and said, "Thank you, Timmy and Timmy's mom." Once upon a time, there was a little girl named Lily. She loved to play outside in
Generated 199 tokens                 
Time for inference 1: 2.0419 sec total                 
Time to first token: 0.1196 sec with parallel prefill.                

      Total throughput: 97.9489 tokens/sec, 0.0102 s/token                 
First token throughput: 8.3582 tokens/sec, 0.1196 s/token                 
 Next token throughput: 103.5252 tokens/sec, 0.0097 s/token                     

Bandwidth achieved: 4.78 GB/s

Warning: Excluding compile in calculations                 
      Average tokens/sec (total): 97.95                 
Average tokens/sec (first token): 8.36                 
Average tokens/sec (next tokens): 103.53 
                
mgschwind@mgschwind-mlt torchchat % python3 torchchat.py export stories15M --quant torchchat/quant_config/desktop.json --output-snap stories15-quant.tc 
NumExpr defaulting to 12 threads.
PyTorch version 2.6.0.dev20241218 available.
Unabled to import torchao experimental quant_api with error:  [Errno 2] No such file or directory: '/Users/mgschwind/tc/torchchat/torchao-build/src/ao/torchao/experimental/quant_api.py'
Using device=mps
Loading model...
Time to load model: 0.30 seconds
Quantizing the model with: {'executor': {'accelerator': 'fast'}, 'precision': {'dtype': 'fast16'}}
Time to quantize model: 0.00 seconds
Exporting model using Snapshot to /Users/mgschwind/tc/torchchat/stories15-quant.tc
 
mgschwind@mgschwind-mlt torchchat % python3 torchchat.py generate stories15M --quant torchchat/quant_config/desktop.json --prompt "once upon a time" --snap stories15-quant.tc

NumExpr defaulting to 12 threads.
PyTorch version 2.6.0.dev20241218 available.
Unabled to import torchao experimental quant_api with error:  [Errno 2] No such file or directory: '/Users/mgschwind/tc/torchchat/torchao-build/src/ao/torchao/experimental/quant_api.py'
Using device=mps 
Loading model...
Time to load model: 0.42 seconds
once upon a time, there was a boy called Jack. He was three years old and very excited. One day, when he was playing in the park, a nosy person came up to him and started pointing at something.
Jack was very curious and he stopped to stare. He saw a big pond, and he wondered what it was. Suddenly, a frog jumped out of the pond and winked at Jack. He was very surprised and the two of them started talking.
“Hey buddy, why are you in the pond? I won't hurt you,” said the frog.
Jack smiled at the frog and asked why he was there. The frog explained that he was looking for some food to feed himself. Jack told the frog that if he helped the frog, he could become friends with him and they could play together in the park.
The frog agreed and Jack carried him out of the p
Generated 199 tokens                 
Time for inference 1: 1.9778 sec total                 
Time to first token: 0.1744 sec with parallel prefill.                

      Total throughput: 101.1238 tokens/sec, 0.0099 s/token                 
First token throughput: 5.7332 tokens/sec, 0.1744 s/token                 
 Next token throughput: 110.3501 tokens/sec, 0.0091 s/token                     

Bandwidth achieved: 17.49 GB/s

Warning: Excluding compile in calculations                 
      Average tokens/sec (total): 101.12                 
Average tokens/sec (first token): 5.73                 
Average tokens/sec (next tokens): 110.35 
                
```

### DIFF
--- a/torchchat/cli/builder.py
+++ b/torchchat/cli/builder.py
@@ -56,6 +56,7 @@ class BuilderArgs:
     gguf_kwargs: Optional[Dict[str, Any]] = None
     dso_path: Optional[Union[Path, str]] = None
     aoti_package_path: Optional[Union[Path, str]] = None
+    snapshot_path: Optional[Union[Path, str]] = None
     pte_path: Optional[Union[Path, str]] = None
     device: Optional[str] = None
     precision: torch.dtype = torch.float32
@@ -81,6 +82,7 @@ class BuilderArgs:
             or (self.dso_path and Path(self.dso_path).is_file())
             or (self.aoti_package_path and Path(self.aoti_package_path).is_file())
             or (self.pte_path and Path(self.pte_path).is_file())
+            or (self.snapshot_path and Path(self.snapshot_path).is_file())
         ):
             raise RuntimeError(
                 "need to specify a valid checkpoint path, checkpoint dir, gguf path, DSO path, AOTI PACKAGE or PTE path"
@@ -136,6 +138,7 @@ class BuilderArgs:
         dso_path = getattr(args, "dso_path", None)
         pte_path = getattr(args, "pte_path", None)
         aoti_package_path = getattr(args, "aoti_package_path", None)
+        snapshot_path = getattr(args, "snapshot_path", None)
 
         is_chat_model = False
         if args.is_chat_model:
@@ -163,6 +166,7 @@ class BuilderArgs:
         output_pte_path = getattr(args, "output_pte_path", None)
         output_aoti_package_path = getattr(args, "output_aoti_package_path", None)
         output_dso_path = getattr(args, "output_dso_path", None)
+        output_snapshot_path = getattr(args, "output_snapshot_path", None)
         if output_pte_path and args.dtype.startswith("fast"):
             if args.dtype == "fast":
                 # As per Kimish, float32 should be faster on ET XNNPACK
@@ -189,6 +193,7 @@ class BuilderArgs:
             dso_path=dso_path,
             aoti_package_path=aoti_package_path,
             pte_path=pte_path,
+            snapshot_path=snapshot_path,
             device=args.device,
             precision=dtype,
             setup_caches=(
@@ -614,6 +619,33 @@ def _initialize_model(
             model = PTEModel(config, builder_args.pte_path)
         except Exception:
             raise RuntimeError(f"Failed to load ET compiled {builder_args.pte_path}")
+    elif builder_args.snapshot_path:
+        # Resolve ModelArgs for constructing the PTEModel
+        # If a manual params_path is provided, use that
+        if builder_args.params_path:
+            config: ModelArgs = ModelArgs.from_params(builder_args.params_path)
+        else:
+            # TODO: Instead of loading the whole model, refactor to call a
+            # helper that generate just model.config
+            with measure_time("Time to load model: {time:.02f} seconds"):
+                model = _load_model(builder_args)
+                device_sync(device=builder_args.device)
+                config = model.config
+                model = None
+        try:
+            model = torch.load(builder_args.snapshot_path, weights_only=False)
+        except Exception:
+            raise RuntimeError(f"Failed to load torchchat snapshot {builder_args.snapshot_path}")
+        # _active_backend() does not allow DSO & AOTI to be true. 
+        # Choose either.
+        set_backend (dso=True, pte=False, aoti_package=False)
+        if (model.config != config):
+            raise RuntimeError("loaded model architecture mismatch")
+        ##        
+        ## import all libraries with custom kernels ans custom operators
+        ## that quantize may be pulling in
+        ##
+
     elif builder_args.distributed:
         pp_degree = builder_args.pp
         tp_degree = builder_args.tp
--- a/torchchat/cli/builder.py
+++ b/torchchat/cli/builder.py
@@ -638,6 +638,7 @@ def _initialize_model(
             raise RuntimeError(f"Failed to load torchchat snapshot {builder_args.snapshot_path}")
         # _active_backend() does not allow DSO & AOTI to be true. 
         # Choose either.
+        from torchchat.utils.build_utils import set_backend
         set_backend (dso=True, pte=False, aoti_package=False)
         if (model.config != config):
             raise RuntimeError("loaded model architecture mismatch")
--- a/torchchat/export.py
+++ b/torchchat/export.py
@@ -29,6 +29,31 @@ default_device = "cpu"
 
 
 """
+Export Snapshot
+"""
+
+
+def export_snapshot(
+    model: nn.Module,
+    device: Optional[str] = None,
+    output_path: str = "model-snapshot.tc",
+) -> str:
+    """
+    Export the model as snapshot.
+
+    Args:
+        model: The model to be exported.
+        device: The device to run the model on.
+        output_path: The path to save the exported model.
+    Returns:
+        The path to the exported model.
+    """
+    assert output_path.endswith(".tc"), "use .tc extension for snapshots"
+    torch.save(model, output_path)
+    return output_path
+
+
+"""
 Export for Server
 """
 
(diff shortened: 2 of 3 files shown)


mikekgfb:
The snapshot load path may need some python imports to pull in all quantization custom ops and custom kernels that quantize.py may make available, so that they are available when a model snapshot gets reloaded.  The best way may be to wholesale import all of them, rather than saving additional info in the snapshot and doing selective import, because there just aren't enough custom ops for wholesale import on reloading a snapshot prohibitive

torchchat/cli/builder.py
```diff
@@ -631,6 +636,34 @@ def do_nothing(max_batch_size, max_seq_length):
...
+        # Resolve ModelArgs for constructing the PTEModel
+        # If a manual params_path is provided, use that
+        if builder_args.params_path:
+            config: ModelArgs = ModelArgs.from_params(builder_args.params_path)
+        else:
+            # TODO: Instead of loading the whole model, refactor to call a
+            # helper that generate just model.config
+            with measure_time("Time to load model: {time:.02f} seconds"):
+                model = _load_model(builder_args)
+                device_sync(device=builder_args.device)
```
Jack-Khuu, line 649: Q: Does the saved artifact still work if the device has changed? I recall this being an issue with AO (one of the reasons why we didn't add saving earlier)

torchchat/export.py
```diff
@@ -28,6 +28,31 @@
...
 
 
+"""
+Export Snapshot
+"""
+
+
+def export_snapshot(
+    model: nn.Module,
+    device: Optional[str] = None,
```
Jack-Khuu, line 38: unused arg?

torchchat/export.py
```diff
@@ -373,14 +399,15 @@ def main(args):
...
+    output_snapshot_path = args.output_snapshot_path
     output_aoti_package_path = args.output_aoti_package_path
 
     if output_pte_path and builder_args.device != "cpu":
         print(
             f"Warning! ExecuTorch export target is controlled by export recipe, not device setting. Ignoring device={builder_args.device} setting."
         )
         builder_args.device = "cpu"
-    elif "mps" in builder_args.device:
+    elif (output_pte_path or output_dso_path or output_aoti_package_path) and "mps" in builder_args.device:
```
Jack-Khuu, line 410: Nice fix

torchchat/export.py
```diff
@@ -28,6 +28,31 @@
...
+
+    Args:
+        model: The model to be exported.
+        device: The device to run the model on.
+        output_path: The path to save the exported model.
+    Returns:
+        The path to the exported model.
+    """
+    assert output_path.endswith(".tc"), "use .tc extension for snapshots"
+    torch.save(model, output_path)
```
Jack-Khuu, line 52: Do we need the whole model or can we get away with just the state_dict? https://github.com/pytorch/torchchat/pull/1280/files

That said if we go with the slimmer state_dict, that's dependent on migration to the AO quant that supports this saving

torchchat/cli/builder.py
mikekgfb: No it might not. Most likely.  It depends really on what quantizations are performed and whether they're implemented on the multiple platforms, and in the same way.  I.e., if it's the same pytorch/python code, for doing computation with quantized numbers and the same quantization formats are supported, then yes.  

If it's a C/C++/CUDA operator, it needs to be supported, with the same name, or with a suitable if/then/else (i.e., don't bake. the "device" setting in)

Quantization weight format layouts need to be consistent, or the loader needs to repack them at load time.  (This is totally plausible to do, but I don't think we do that today.  I think in the 4b case ""we just know".  I tried to change that, but the need/priority wasn't similarly perceived by everybody.)

If it's saved, and reloaded, most (all?) decisions you made are set in stone, like quantization schemes etc (Otherwise, you'd be loading from scratch?). In some sense that's similar to how dso/aoti/pte-output-path / load-dso/aoti/pte-path work, and that's why it's modeled after that export and reload facility.  You don't get to change the PTE target on reload, or the device that an aoti model has been compiled for.

torchchat/export.py
mikekgfb: It's a cut and paste thing, and wanted to keep args consistent.  Mind you, we could put the device in the file or some other such, and check on reload that it's the same.  (I think MPS/CPU are sorta fungible, which might help on Macs with quantization when you run out of kernel memory to quantize large models.  CPU could use paging to ssd for that. eg discussion on #1483)

torchchat/export.py
mikekgfb: You rewrite the code as part of the quantization.  If you don't save the code, then you must exactly replicate what quantization options you used, create an empty model, quantize it, and then load the state dict over it.  Either you transfer the whole responsibility on the user (good luck, you'll die the death of many cuts when users make mistakes and complain that this facility does not work), or you need to save an ungodly amount of information about options used for the original quantization process.

mikekgfb:
> Love the PR. This has been on the backlog for a bit (and unblocks a potential path to quantize at download time + discard the unquantized snapshot)
> 
> Left some comments

you're welcome :) Glad you like it!


PS: If you quantize at download, you need to have a way to specify quantization options.  Totally doable, but not sure about discarding the unquantized snapshot.  As a user, if I change my mind, or forgot device, and if I want to change quantization, it has to re-download.  Even on my corpnet in Silicon Valley, the larger models are painful to download -- even worse for the rest of the world. But then again, I have a fairly large disk on my laptop (but on the tradeoff of download bw vs ssd size, i think ssd size wins for most? 

Then again, if the user thinks s/he might change her mind, she should just download and quantize separately...

Jack-Khuu:
Definitely want to spin up a RFC before we push the quantized download through

> If you quantize at download, you need to have a way to specify quantization options... As a user, if I change my mind, or forgot device, and if I want to change quantization, it has to re-download. 

Our reference here is Ollama which gives a "it just works" experience due to their default to quantized inference (GGUF format aside). We'd give a default quantization setting when downloading in this fashion.

> Not sure about discarding the unquantized snapshot...
> If the user thinks s/he might change her mind, she should just download and quantize separately

This will always be an option :D 
And if we do sell things right will be something that users gravitate to (or they point to a pre-downloaded snapshot)

torchchat/cli/builder.py
Jack-Khuu: I think expecting the exporting conditions to be the the same as the executing conditions is a fair start

torchchat/export.py
Jack-Khuu: We can update the serialization logic as the AO migration finishes (i.e. this PR is good), but I'm not sure that's the case anymore with AO. I was under the impression that the model itself is unaffected and that only the weights are changed

https://github.com/pytorch/ao/blob/48fdd310b3977a0db2ceba37a7725192cd2aafd4/docs/source/serialization.rst#L62 

cc: @vmpuri @HDCharles

Jack-Khuu (approved):

torchchat/export.py
HDCharles: you should be able to get by with just the state dict, there are a few apis that don't work that way, but all the subclass ones do. Thats like 75% of the reason that we went with subclasses instead of module swaps.

jack, you're link is a good resource, an alternate reference are our serialization tests to see what is explicitly tests

see e.g. https://github.com/pytorch/ao/blob/48fdd310b3977a0db2ceba37a7725192cd2aafd4/test/integration/test_integration.py#L1322-L1334

PR #1350, opened 2024-11-07
Create run-readme-pr-linuxaarch64

mikekgfb:
Test torchchat on aarch64 linux

PR #1469, opened 2025-01-22
Update run-readme-pr-linuxaarch64.yml to use correct runner

Jack-Khuu:
https://github.com/pytorch/torchchat/pull/1350 used `linux-aarch64` as the runner when we should be using `linux.arm64.2xlarge` for aarch64 instead

### DIFF
--- a/.github/workflows/run-readme-pr-linuxaarch64.yml
+++ b/.github/workflows/run-readme-pr-linuxaarch64.yml
@@ -10,6 +10,9 @@ on:
 jobs:
   test-readme-cpu:
     uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    permissions:
+      id-token: write
+      contents: read
     with:
       runner: linux.arm64.m7g.4xlarge
       docker-image: "pytorch/manylinux2_28_aarch64-builder:cuda12.6"
@@ -35,6 +38,9 @@ jobs:
 
   test-quantization-cpu:
     uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    permissions:
+      id-token: write
+      contents: read
     with:
       runner: linux.arm64.2xlarge
       docker-image: "pytorch/manylinux2_28_aarch64-builder:cpu-aarch64-main"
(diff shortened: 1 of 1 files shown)


.github/workflows/run-readme-pr-linuxaarch64.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   test-readme-cpu:
     uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
     with:
-      runner: linux-aarch64
+      runner: linux.arm64.m7g.4xlarge
       gpu-arch-type: cuda
       gpu-arch-version: "12.1"
       timeout: 60
```
atalman, line 17: try passing 
```
docker-image: "pytorch/manylinuxaarch64-builder:cuda12.1-main"
```

Jack-Khuu:
fyi: @mikekgfb we looking into it

.github/workflows/run-readme-pr-linuxaarch64.yml
Jack-Khuu: Doesn't seem like it can find the Docker-image verbatim

Testing with the 12.6 version found in [pt/pt](https://github.com/pytorch/pytorch/blob/0d28188cc8fa8b895d897ca31171223268d30a7f/.github/scripts/generate_binary_build_matrix.py#L175)

atalman: If using linux_job_v2.yml you can try using latest image ``pytorch/manylinux2_28_aarch64-builder:cuda12.6``

Jack-Khuu: Doesn't look like the cuda version is there `manylinux2_28_aarch64-builder:cuda12.6`, but the CPU variant `:cpu-aarch64-main` with `linux_job_v2` seems to be the right track

 Now we're just down to missing devtoolset-10-binutils, which is curious since pt/pt uses v10 for aarch64
 ```
No match for argument: devtoolset-10-binutils
  Error: Unable to find a match: devtoolset-10-binutils
```

Gasoonjia (approved):

PR #1473, opened 2025-01-23
Update Android NDK version to r27c

Jack-Khuu:


Jack-Khuu:
Yup, we should add a test for the demos; might be able to pull it from the ET repo

The current test we're trying to fix just builds the android runner

mikekgfb:
@Jack-Khuu maybe this? https://github.com/pytorch/executorch/issues/5387#issuecomment-2353323570 solves the -lextension_tensor ?

Jack-Khuu:
This is for the Python build so it's not directly related

That said I think I found the culprit: https://github.com/pytorch/executorch/pull/7473
- XNNPack pybindings are now on by default

mikekgfb:
> This is for the Python build so it's not directly related
> 
> That said I think I found the culprit: [pytorch/executorch#7473](https://github.com/pytorch/executorch/pull/7473)
> 
> * XNNPack pybindings are now on by default

If that doesn't do it, maybe you want to add environment variables here.  Because clearly that's how we turn on these extensions for other extensions, and the traceback of the build shows the EXTENSION TENSOR library as "OFF" atm.

https://github.com/pytorch/torchchat/blob/main/runner/build_android.sh#L24-L25

Jack-Khuu:
Yup, yup that was the log that tipped me off that something changed in the ET installs

PR #1462, opened 2025-01-16
Increment start_pos by encoded size in generate

nlpfollower:
Hi, seems like the size of `encoded` should contribute to `start_pos` on each turn. Without this change, I'm getting inconsistent generations, as the model looks up kv from the middle of the cache, rather than the end. This also feels like a natural place to increment `start_pos`, since right after we also increment by the size of generated `token_tensor`.

PR #1474, opened 2025-01-23
Decode_n_tokens yield eos/eot token

nlpfollower:
Seems like we should yield the `cur_token`/`next_token` when it's an eos or eot so that it is included in the `generator_func` in `chat`. Also this way `start_pos` is incremented to an empty position in the cache. Repro similar to #1462, although there is no visible response degradation. 

I've pushed logs from my trace (https://github.com/nlpfollower/torchchat/pull/2). You can see how without this yield the first id in the `input_pos` at the beginning of a round overlaps, and overwrites, the last cache entry made in the previous turn.


Jack-Khuu:
Thanks for another fix; I'll try to verify over the weekend

Jack-Khuu (approved):
My guts says this is legit, I'll give it another look

PR #1475, opened 2025-01-24
Explicitly turning off pybindings for ExecuTorch install unless requested

Jack-Khuu:
ExecuTorch now has XNN pybinding built by default https://github.com/pytorch/executorch/pull/7473

Previously it was not built by default


Gasoonjia (approved):

mikekgfb:
👍

mergennachin:
Thanks @Jack-Khuu, didn't realize this. 

Also, while we're at it. 

We have two commands: install_requirements (only installs required dependency) and install_executorch (which does install_requirements underneath the hood as well as installs executorch).

See attached PRs to this issue (https://github.com/pytorch/executorch/issues/7730)

Issue #1472, opened 2025-01-22
Numpy Version Pin Bump: >= 2.0


Jack-Khuu:
### 🐛 Describe the bug

ExecuTorch recently bumped their numpy requirements to numpy == 2.0 in https://github.com/pytorch/executorch/commit/a7b5297f95101c32b9e45fd1ee3e5dfb4a00c96b
This puts torchchat in a finicky spot since the current requirements are tied to under 2.0 due to GGUF support requiring < 2.0 (see blame for previous attempts)
https://github.com/pytorch/torchchat/blob/fb65b8bbc405066162fbae1b060eb507687ebabf/install/requirements.txt#L19-L20

While not actively an issue, as soon as an ExecuTorch pin bump is required, this will become a hard blocker.

**Task:** Make a requirements version pinbump `numpy >= 2.0`

### Versions

N/A

Jack-Khuu:
fyi: @Gasoonjia, @vmpuri, @mergennachin (ET)

Jack-Khuu:
> where does it fail when the version is 2 or above?

It was implicit dependency in llama.cpp; the project was using interfaces that were deprecated in numpy 2+
That said looks like it got fixed in December (https://github.com/ggerganov/llama.cpp/pull/9772)

If we're able to just drop it and things work, that's the best case scenario

mergennachin:
yeah, let's just try updating this commit 

https://github.com/pytorch/torchchat/blob/main/.github/workflows/pull.yml#L729

PR #1479, opened 2025-01-24
Update numpy requirements to no longer upper bound on 2.0

Jack-Khuu:
https://github.com/pytorch/torchchat/issues/1472

### DIFF
--- a/install/requirements.txt
+++ b/install/requirements.txt
@@ -16,8 +16,7 @@ jinja2
 # Miscellaneous
 snakeviz
 sentencepiece
-# numpy version range required by GGUF util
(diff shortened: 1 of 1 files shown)


Gasoonjia (approved):

mergennachin:
@Jack-Khuu 

You don't need to update this commit hash? https://github.com/pytorch/torchchat/blob/main/.github/workflows/pull.yml#L729

to include the fix from llama.cpp?

Jack-Khuu:
Forgot about that, good catch

None of the tests flagging errors is concerning too... cc @Gasoonjia

PR #1476, opened 2025-01-24
source instead of exec in run-readme-pr-macos.yml

mikekgfb:
source test commands instead of executing them.  
(Possible fix for #1315 )

PR #1489, opened 2025-02-06
use template from pull.yml for run-readme-pr.yml since pip3 not available in current environment

mikekgfb:
use template from pull.yml for run-readme-pr.yml since pip3 not available in current environment


mikekgfb:
Resolved by #1476

PR #1503, opened 2025-02-24
Add XPU support for AOT inductor

jenniew:
Add XPU support for AOT inductor.


install/install_requirements.sh
```diff
@@ -115,6 +115,17 @@ fi
     "${REQUIREMENTS_TO_INSTALL[@]}"
 )
 
+# Temporatory instal torchtune nightly from cpu nightly link since no torchtune nightly for xpu now
+# TODO: Change to install torchtune from xpu nightly link, once torchtune xpu nightly is ready
+if [[ -x "$(command -v xpu-smi)" ]];
+then
+(
+  set -x
```
Jack-Khuu, line 123: Can we reuse the conditional on line 84?

Jack-Khuu (approved):

install/install_requirements.sh
jenniew: No, because line 84 is to install using pytorch xpu link: https://download.pytorch.org/whl/nightly/xpu. Since we don't have torchtune nightly for xpu now, we temporarily to install torchtune cpu nightly from cpu link: https://download.pytorch.org/whl/nightly/cpu. After xpu nightly is ready, we'll update line 84 to install from pytorch xpu link.

Jack-Khuu: Ah I meant reuse the `if [[ -x "$(command -v xpu-smi)" ]];` check

This PR is fine to merge as-is though since torchtune nightly is going to remove this anyways 😃

PR #1510, opened 2025-03-13
Move tokenizer info into pte for et

silverguo:
```
$python3 torchchat.py export llama3.1 --quantize torchchat/quant_config/mobile.json --output-pte-path exportedModels/llama3.1.pte
....
The methods are:  {'forward'}
The config methods are:  {'tokenizer_type'}
....
$torchchat/utils/scripts/build_native.sh et
....
$cmake-out/et_run exportedModels/llama3.1.pte -z `python3 torchchat.py where llama3.1`/tokenizer.model -l 3 -i "Once upon a time"
....
num_methods: 2
err: 0
status: 0
....
```
It seems that I can load that config method in `program` but didn't figure out a way to get the output.


runner/run.cpp
```diff
@@ -867,6 +873,37 @@ int main(int argc, char *argv[]) {
...
+  }
+  HierarchicalAllocator planned_memory(
+      {planned_spans.data(), planned_spans.size()});
+  MemoryManager memory_manager(&method_allocator, &planned_memory);
+  auto tokenizer_method = transformer.runner->program()->load_method("tokenizer_type", &memory_manager);
+
+  Error err = tokenizer_method->execute();
+  fprintf(stderr, "err: %u\n", err);
+
+  std::vector<EValue> outputs(tokenizer_method->outputs_size());
```
Jack-Khuu, line 903: Can you print outputs?

runner/run.cpp
```diff
@@ -867,6 +873,37 @@ int main(int argc, char *argv[]) {
...
+  for (size_t id = 0; id < num_memory_planned_buffers; ++id) {
+    // .get() will always succeed because id < num_memory_planned_buffers.
+    size_t buffer_size =
+        static_cast<size_t>(method_meta->memory_planned_buffer_size(id).get());
+    planned_buffers.push_back(std::make_unique<uint8_t[]>(buffer_size));
+    planned_spans.push_back({planned_buffers.back().get(), buffer_size});
+  }
+  HierarchicalAllocator planned_memory(
+      {planned_spans.data(), planned_spans.size()});
+  MemoryManager memory_manager(&method_allocator, &planned_memory);
```
Jack-Khuu, line 897: Can you try `nullptr` for planned_memory? We shouldn't need planned memory if we're not hunting for a tensor

Jack-Khuu (approved):

Issue #1504, opened 2025-02-25
Update CI Jobs in anticipation for Cuda 12.4 deprecation


Jack-Khuu:
### 🚀 The feature, motivation and pitch

PT is deprecating CUDA 12.4 nightly builds: https://github.com/pytorch/test-infra/pull/6333

This repo will need to update CI and testing in response to this

HonestDeng:
Hi. I want to solve it and to be a long-time contributor. Is there any procedure I need go through? Or is there any help? Thanks

Jack-Khuu:
Turns out we don't need to bump the cuda configs in the ci per https://github.com/pytorch/test-infra/issues/6465

PR #1515, opened 2025-03-24
Update CI Jobs in anticipation for Cuda 12.4 deprecation

HonestDeng:
This PR aimed to solve this [issue](https://github.com/pytorch/torchchat/issues/1504).

Need to update ci configs and install_requirements.


Jack-Khuu:
Thanks for the attempt @HonestDeng, as per the comment in test-infra, turns out we didn't need to make a CI change

PR #1523, opened 2025-04-07
Enable compile on MPS with experimental warning

manuelcandales:
This enables compile on MPS, modifying the warning message to say this is experimental and is not ready for broad use yet
We also adjust inductor's config to avoid the inductor's decomposition of mm if running on MPS backend

```
python3 torchchat.py generate stories15M --dtype float16 --prompt "Once upon a time" --num-samples 3 --device mps --compile
```

### DIFF
--- a/torchchat/cli/cli.py
+++ b/torchchat/cli/cli.py
@@ -579,13 +579,11 @@ def arg_init(args):
                 print(f'overriding json-specified device {executor_handler["accelerator"]} with cli device {args.device}')
                 executor_handler["accelerator"] = args.device
 
-    # if "mps" in args.device:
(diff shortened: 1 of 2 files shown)


Jack-Khuu:
We added this for MPS LLama 3.2-Vision. Does this affect your work here?

https://github.com/pytorch/torchchat/blob/2640f6aef5a22aa4c48e303a0c5de38640d2d8db/torchchat.py#L15

Jack-Khuu (approved):

manuelcandales:
> PYTORCH_ENABLE_MPS_FALLBACK

I don't think so. Will experiment with it, but shouldn't have any effect on stories or text only llamas

PR #1532, opened 2025-04-21
decode_n_tokens clean up

manuelcandales:
Deletes unused python arrays

### DIFF
--- a/torchchat/generate.py
+++ b/torchchat/generate.py
@@ -535,7 +535,6 @@ class LocalGenerator:
         attention_backend: SDPBackend = torch.nn.attention.SDPBackend.MATH,
         **sampling_kwargs,
     ):
-        new_tokens, new_probs = [], []
         encountered_eos = False
         for _i in range(
             num_new_tokens - 1
@@ -553,12 +552,10 @@ class LocalGenerator:
                     **sampling_kwargs,
                 )
                 input_pos += 1
-                new_tokens.append(next_token.clone())
-                callback(new_tokens[-1], done_generating=_i == num_new_tokens - 2)
-                if need_probs or next_prob is None:
+                callback(next_token.clone(), done_generating=_i == num_new_tokens - 2)
+                if not need_probs or next_prob is None:
                     yield out_token, None
                 else:
-                    new_probs.append(next_prob.clone())
                     yield out_token, next_prob.clone()
                 cur_token = next_token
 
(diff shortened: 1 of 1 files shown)


torchchat/generate.py
```diff
@@ -535,7 +535,6 @@ def decode_n_tokens(
         attention_backend: SDPBackend = torch.nn.attention.SDPBackend.MATH,
         **sampling_kwargs,
     ):
-        new_tokens, new_probs = [], []
```
manuelcandales, line 538: Not really used. This function yields individual tokens and probabilities, not the arrays.

torchchat/generate.py
```diff
@@ -553,12 +552,10 @@ def decode_n_tokens(
                     **sampling_kwargs,
                 )
                 input_pos += 1
-                new_tokens.append(next_token.clone())
-                callback(new_tokens[-1], done_generating=_i == num_new_tokens - 2)
-                if need_probs or next_prob is None:
```
manuelcandales, line 558: This was backwards. It should be: `if not need probs or next_prob is None:`. Otherwise you are saying, if you need the probabilities you are getting None, and if you don't need them you are getting them.

torchchat/generate.py
```diff
@@ -788,7 +784,6 @@ def generate(
                 input_pos = input_pos + num_added
                 next_token = next_tokens[-1]
         else:
-            generated_tokens = []
```
manuelcandales, line 791: Not used. Generated tokens are appended after the call to `decode_n_tokens`, but then nothing happens.

torchchat/generate.py
```diff
@@ -553,12 +552,10 @@ def decode_n_tokens(
                     **sampling_kwargs,
                 )
                 input_pos += 1
-                new_tokens.append(next_token.clone())
-                callback(new_tokens[-1], done_generating=_i == num_new_tokens - 2)
-                if need_probs or next_prob is None:
+                callback(next_token.clone(), done_generating=_i == num_new_tokens - 2)
+                if not need_probs or next_prob is None:
```
Jack-Khuu, line 556: Everything else is just cleaning up unused code

`not need_probs` is the only real change and in the non-speculative path is always false so the old check is effectively just `if next_prob is None`
https://github.com/pytorch/torchchat/blob/701d82658d0a48799131a7f6692574648068f118/torchchat/generate.py#L799

I think we should drop the check instead of negating here
```suggestion
                if next_prob is None:
```

Jack-Khuu (approved):

torchchat/generate.py
Jack-Khuu: Honestly this is a nit we can  just merge

PR #1534, opened 2025-04-22
Accumulate tokens in generate mode

manuelcandales:
We accumulate tokens in generate mode before calling callback on them.
This avoids synchronizing GPU and CPU on each token, improving performance on MPS backend.

### DIFF
--- a/torchchat/generate.py
+++ b/torchchat/generate.py
@@ -230,6 +230,7 @@ class GeneratorArgs:
     max_autotune: bool = False
     # (Misnomer) See Issue: https://github.com/pytorch/torchchat/issues/1273
     is_torchtune_model: bool = False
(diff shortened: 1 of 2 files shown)


torchchat/generate.py
```diff
@@ -294,6 +295,7 @@ def from_args(cls, args):
             sequential_prefill=sequential_prefill,
             max_autotune=args.max_autotune,
             is_torchtune_model=args.model and args.model.endswith("tune"),
+            accumulate_tokens=getattr(args, "accumulate_tokens", 8),
```
Jack-Khuu, line 298: Unrelated to this PR: I should fix this so we don't have duplicate defaults

Jack-Khuu (approved):

Issue #1518, opened 2025-03-31
Simplify TokenizerArgs __post_init__: Unnecessarily verbose


Jack-Khuu:
### 🚀 The feature, motivation and pitch

`TokenizerArgs.__post_init__` has grown quite verbose/redundant and could use a bit of simplification

https://github.com/pytorch/torchchat/blob/1384f7d3d7af0847d8364fe7b300a8b49f2213c2/torchchat/cli/builder.py#L244-L289

Task: Simplify the logic in __post_init__ to reduce redundancy
---
To test, run a model with each tokenizer type:
- python torchchat.py generate llama2
- python torchchat.py generate llama3
- python torchchat.py generate granite-code

mu-lippy:
I'd like to work on this, please.

Jack-Khuu:
Give it a shot 😃 

Also if you haven't already, definitely join the discord https://discord.com/invite/hm2Keduk3v

mu-lippy:
I'm sorry if this question is silly, but do I need to write tests after I make changes? Thanks!

Jack-Khuu:
Usually yes, but in this case the integration tests should catch it so you should be good without (bad tokenizer casting fails fast)

mu-lippy:
I think dealing with the different parameter naming conventions for each tokenizer makes this hard to deal with. When I try to make it more concise, it looks a lot uglier and harder to read. I may be wrong, but it might be better to leave this method as is?
If not, I'd love to see a more experienced coder's solution.

Jack-Khuu:
I can give a hint:
* For this task we don't need to dive deep into the tokenizers themselves
* At most one of these fields are set to true: `{self.is_tiktoken, self.is_sentencepiece, self.is_hf_tokenizer } `. Let's replace this arg with an enum to make it more legible

mu-lippy:
I think this might be beyond me - it looks like I will have to make changes to other parts of this file besides the TokenizerArgs class, and I'm not confident I can do this without breaking it. I'm sorry

Jack-Khuu:
No worries, I'll put this back up for grabs

If you're still interested, feel free to put up an initial attempt and we can work through it :)

PR #1535, opened 2025-04-25
Simplify `TokenizerArgs.__post_init__` with Enum Tokenizer Type

zhenyan-zhang-meta:
Summary:
Simplify `TokenizerArgs.__post_init__` with enum tokenizer type, since only one of the tokenizer type can be true.

We want to touch as less code outside of `__post_init__` as possible at the moment.

Test Plan:
python torchchat.py generate llama2|llama3|granite-code

Reviewers:
@Jack-Khuu

Subscribers:

Issue:
https://github.com/pytorch/torchchat/issues/1518

### DIFF
--- a/torchchat/cli/builder.py
+++ b/torchchat/cli/builder.py
@@ -8,6 +8,7 @@ import argparse
 import os
 import sys
 from dataclasses import dataclass
+from enum import Enum
 from pathlib import Path
 from typing import Any, Dict, Optional, Tuple, Union
 
@@ -237,13 +238,16 @@ class BuilderArgs:
         speculative_builder_args.pte_path = None
         return speculative_builder_args
 
+class TokenizerType(Enum):
+    NONE = 0
+    TIKTOKEN = 1
+    SENTENCEPIECE = 2
+    HF_TOKENIZER = 3
 
 @dataclass
 class TokenizerArgs:
     tokenizer_path: Optional[Union[Path, str]] = None
-    is_sentencepiece: bool = False
-    is_tiktoken: bool = False
-    is_hf_tokenizer: bool = False
+    tokenizer_type: TokenizerType = TokenizerType.NONE
     t: Optional[Any] = None
 
     def __post_init__(self):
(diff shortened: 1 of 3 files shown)


torchchat/cli/builder.py
```diff
@@ -294,12 +299,14 @@ def validate_model(
         if model is None:
             return
 
-        if sum([self.is_tiktoken, self.is_hf_tokenizer, self.is_sentencepiece]) != 1:
+
+        is_tiktoken = self.is_tiktoken()
+        is_sentencepiece = self.is_sentencepiece()
+        is_hf_tokenizer = self.is_hf_tokenizer()
+
+        if sum([is_tiktoken, is_hf_tokenizer, is_sentencepiece]) != 1:
```
Jack-Khuu, line 307: We can replace this by just checking if the tokenizer enum is None

torchchat/cli/builder.py
```diff
@@ -273,19 +273,24 @@ def __post_init__(self):
...
+            self.tokenizer_type = TokenizerType.HF_TOKENIZER
             return
         except:
             pass
 
-        self.is_tiktoken = False
-        self.is_sentencepiece = False
-        self.is_hf_tokenizer = False
+        self.tokenizer_type = TokenizerType.NONE
         self.t = None
```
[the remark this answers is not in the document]
Jack-Khuu: We can actually drop all the logic here after the HF tokenizer check, tokenizer_type and .t are already set to these by default

torchchat/cli/builder.py
```diff
@@ -273,19 +273,29 @@ def __post_init__(self):
...
+            self.tokenizer_type = TokenizerType.HF_TOKENIZER
             return
         except:
             pass
 
-        self.is_tiktoken = False
-        self.is_sentencepiece = False
-        self.is_hf_tokenizer = False
-        self.t = None
         return
```
Jack-Khuu, line 281: nit: return is not needed

torchchat/cli/builder.py
```diff
@@ -294,12 +304,13 @@ def validate_model(
         if model is None:
             return
 
-        if sum([self.is_tiktoken, self.is_hf_tokenizer, self.is_sentencepiece]) != 1:
+        if self.is_tokenizer_none():
```
Jack-Khuu, line 307: Let's keep it simple and check for `self.tokenizer_type == TokenizerType.NONE` here

We can skip the the assert/helper

Jack-Khuu (approved):
Looks good, fix the nits then go ahead and merge

PR #1537, opened 2025-05-01
Improve Quantization Setup Prompts and Doc

zhenyan-zhang-meta:
Currently we don't prompt people to clone torchao first before running `bash torchchat/utils/scripts/build_torchao_ops.sh`. We can improve the prompts and doc similar to what we already have for `bash torchchat/utils/scripts/build_native.sh et link_torchao_ops`.

cc @Jack-Khuu @byjlw

### DIFF
--- a/torchchat/utils/scripts/build_torchao_ops.sh
+++ b/torchchat/utils/scripts/build_torchao_ops.sh
@@ -12,13 +12,13 @@ if [[ "$device" != "cpu" && "$device" != "mps" ]]; then
   exit 1
 fi
 
+source "$(dirname "${BASH_SOURCE[0]}")/install_utils.sh"
+
 if [ ! -d "${TORCHCHAT_ROOT}/torchao-build" ]; then
   echo "Directory ${TORCHCHAT_ROOT}/torchao-build does not exist."
   echo "Make sure you run bash torchchat/utils/scripts/clone_torchao.sh"
   exit 1
 fi
-
-source "$(dirname "${BASH_SOURCE[0]}")/install_utils.sh"
 
 pushd ${TORCHCHAT_ROOT}
 find_cmake_prefix_path


docs/quantization.md
```diff
@@ -149,6 +149,7 @@ From the torchchat root directory, run
 bash torchchat/utils/scripts/build_torchao_ops.sh
 ```
 
+Note that before running it, you must first clone torchao with `bash torchchat/utils/scripts/clone_torchao.sh` if you have not done so already.
```
Jack-Khuu, line 152: `clone_torchao` is called as part of the install_requirements, so it might be better for us to remind users to call that first

Jack-Khuu (approved):
Thanks, mini nit good good to merge

docs/quantization.md
zhenyan-zhang-meta: Do you mean `./install/install_requirements.sh`, I don't see torchao is cloned to the path that `build_torchao_ops.sh` uses.

torchchat/utils/scripts/build_native.sh
```diff
@@ -70,14 +70,14 @@ git submodule sync
...
-    echo "Make sure you run install_executorch_libs"
+    echo "Make sure you run bash torchchat/utils/scripts/install_et.sh"
     exit 1
   fi
 
   if [[ "$LINK_TORCHAO_OPS" == "ON" ]]; then
     if [ ! -d "${TORCHCHAT_ROOT}/torchao-build" ]; then
       echo "Directory ${TORCHCHAT_ROOT}/torchao-build does not exist."
-      echo "Make sure you run clone_torchao"
+      echo "Make sure you run bash torchchat/utils/scripts/clone_torchao.sh"
```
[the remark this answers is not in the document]
zhenyan-zhang-meta: I was thinking that, if we have to let users run this, with `clone_torchao` they might not know how to run it.

PR #1542, opened 2025-05-12
Push torchtune install deeper in generate._gen_model_input and undo import order change for etmodel

Jack-Khuu:
Follow up patch from https://github.com/pytorch/torchchat/pull/1539 which was meant to make torchtune an optional import

---
Tested with:

Success only when torchtune installed
```
python torchchat.py generate llama3.2-11B --prompt "What's in this image?" --image-prompt assets/dog.jpg --num-samples 3
```

Success regardless of torchtune install
```
python3 torchchat.py generate llama3.1-base --prompt "Once upon a time,"
```

### DIFF
--- a/torchchat/generate.py
+++ b/torchchat/generate.py
@@ -869,13 +869,6 @@ class LocalGenerator:
         max_new_tokens: Optional[int] = None,
         max_seq_len: Optional[int] = 2048,
     ) -> Tuple[torch.Tensor, Optional[Dict[str, Any]]]:
-        # torchtune model definition dependencies
-        from torchtune.data import Message, padded_collate_tiled_images_and_mask
-        from torchtune.models.llama3_2_vision._model_builders import (
-            llama3_2_vision_transform,
-        )
-        from torchtune.training import set_default_dtype
-
         """
         Convert prompt and image prompts into consumable model input args.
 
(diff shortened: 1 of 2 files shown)


torchchat/model.py
```diff
@@ -1054,13 +1054,13 @@ def apply_rotary_emb(x: Tensor, freqs_cis: Tensor) -> Tensor:
...
 try:
-    # For llama::sdpa_with_kv_cache.out, preprocess ops
-    from executorch.extension.llm.custom_ops import custom_ops  # no-qa
     from executorch.extension.pybindings import portable_lib as exec_lib
 
     # ET changed the way it's loading the custom ops so it's not included in portable_lib but has to be loaded separately.
     # For quantized_decomposed ops
     from executorch.kernels import quantized  # no-qa
+    # For llama::sdpa_with_kv_cache.out, preprocess ops
+    from executorch.extension.llm.custom_ops import custom_ops  # no-qa
```
zhenyan-zhang-meta, line 1063: nit: what's the difference this change made?

torchchat/generate.py
```diff
@@ -911,6 +904,14 @@ def _gen_model_input(
             return encoded, None
 
         # Llama 3.2 11B
```
zhenyan-zhang-meta, line 906: Are we sure that everything below is for llama 3.2 11b? If so good to go, otherwise we need an if-else.

Will we in the future have more possibilities here in this func? If so, we list them with an if-else or switch-case chain.

zhenyan-zhang-meta (approved):
LGTM with comments. Can land after signal pass.

torchchat/generate.py
Jack-Khuu: yup everything below is 11B. 
If there's more models that require casing this function as a whole should get refactored at that time

torchchat/model.py
Jack-Khuu: This was the original order prior to the other PR
We preserve the order due to how extension libraries interact (we want custom_ops to overwrite potential conflicts)

PR #1549, opened 2025-09-10
Change macOS runner version in workflow

ZainRizvi:
Removing dependency on self-hosted macOS runners, in prep for migration to meta-pytorch org


ZainRizvi:
Mps jobs succeeded. Merging changes

